PR #2, opened 2020-05-12
Add /usr/local/bin to secure_path, Add br_netfilter on centos

St0rmingBr4in:
This fixes a few issues when running the playbook on centos:

 - Missing `br_netfilter` module
 - Missing `/usr/local/bin` in sudo secure_path

Signed-off-by: Julien DOCHE <julien.doche@gmail.com>

### DIFF
--- a/roles/prereq/tasks/main.yml
+++ b/roles/prereq/tasks/main.yml
@@ -18,6 +18,18 @@
     state: present
     reload: yes
 
+- name: Add br_netfilter to /etc/modules-load.d/
+  copy:
+    content: "br_netfilter"
+    dest: /etc/modules-load.d/br_netfilter.conf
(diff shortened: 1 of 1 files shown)


roles/prereq/tasks/main.yml
```diff
@@ -18,6 +18,17 @@
     state: present
     reload: yes
 
+- name: Add br_netfilter to /etc/modules-load.d/
```
itwars, line 21: Do we need to add br_netfilter for all distro, or just Redhat/Centos?

St0rmingBr4in: Good catch, I had to add it only on centos

itwars: Excellent!

PR #14, opened 2020-05-12
fix typo for master_ip

4m3ndy:


geerlingguy:
I read this as `master_pi` and was like "yeah, I'm building a Pi cluster so it's the `master_pi`... but then I realized if you're _not_ building a Pi cluster, then that would make no sense! So in some cases it works either way, but I think this change is clear for _all_ use cases :)

PR #18, opened 2020-05-15
Fixes #17: Fix is changed check so reboot task doesn't fail.

### LINKED ISSUE #17: The conditional check 'boot_cmdline | changed' failed. No filter named 'changed'
In Ansible 2.9.x, the check for `Rebooting on Raspbian` fails with the following error message:

```
fatal: [18.206.98.159]: FAILED! => {"msg": "The conditional check 'boot_cmdline | changed' failed. The error was: template error while templating string: no filter named 'changed'. String: {% if boot_cmdline | changed %} True {% else %} False {% endif %}\n\nThe error appears to be in '/Users/jgeerling/Downloads/youtube-10k-pods/attempt-two-k3s/k3s-ansible/roles/raspbian/tasks/main.yml': line 19, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Rebooting on Raspbian\n  ^ here\n"}
```

The fix is to set the conditional to `boot_cmdline is changed`.

geerlingguy:
Closes #17.﻿

### DIFF
--- a/roles/raspbian/tasks/main.yml
+++ b/roles/raspbian/tasks/main.yml
@@ -19,5 +19,5 @@
 - name: Rebooting on Raspbian
   reboot:
   when:
-    - boot_cmdline | changed
+    - boot_cmdline is changed
     - ansible_facts.architecture is search("arm")


St0rmingBr4in:
LGTM

PR #21, opened 2020-05-16
Add ubuntu on Raspberry pi support

mickkael:


roles/ubuntu/tasks/main.yaml
```diff
@@ -0,0 +1,23 @@
...
+    path: /boot/firmware/cmdline.txt
+  register: cmdline
+
+- name: Activating cgroup on Ubuntu for Raspberry
+  lineinfile:
+    path: /boot/firmware/cmdline.txt
+    regexp: '^(.*rootwait fixrtc)$'
+    line: '\1 cgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory'
+    backrefs: true
+  when: ( cmdline.stat.path is defined )
```
geerlingguy, line 13: This can be simplified to just:

```yaml
  when:
    - cmdline.stat.path is defined
    - ansible_facts.architecture is search("aarch")
```

(as well as in the next task's `when` condition).

roles/ubuntu/tasks/main.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
geerlingguy, line 1: This role doesn't seem to be called from anywhere...?

roles/ubuntu/tasks/main.yaml
```diff
@@ -0,0 +1,23 @@
...
+    regexp: '^(.*rootwait fixrtc)$'
+    line: '\1 cgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory'
+    backrefs: true
+  when: ( cmdline.stat.path is defined )
+          and
+        ( ansible_facts.architecture is search("aarch") )
+  register: boot_cmdline
+
+- name: Rebooting on Ubuntu for Raspberry
+  shell: reboot now
```
geerlingguy, line 19: You can use Ansible's `reboot` module for idempotence and so the playbook waits for the reboot to complete:

```yaml
  reboot:
```

(And then you can drop the `ignore_errors`.)

PR #23, opened 2020-05-17
Fixes #22: Set proper file mode using an octal.

### LINKED ISSUE #22: File mode for download tasks needs to be in octal
For the file-related tasks in the download role, the `mode` should be an octal (or needs to be quoted), otherwise the permissions will not be set properly.

https://github.com/rancher/k3s-ansible/blob/master/roles/download/tasks/main.yml#L8-L15

These should be like:

```
mode: 0755
```

From Ansible's docs on get_url:

> You must either add a leading zero so that Ansible's YAML parser knows it is an octal number (like `0644` or `01777`) or quote it (like `'644'` or `'1777'`) so Ansible receives a string and can do its own conversion from string into number.

It seems this is already correct in the `k3s` role file-related tasks, so let's get these download tasks in the same format.

geerlingguy:
Closes #22.﻿

### DIFF
--- a/roles/download/tasks/main.yml
+++ b/roles/download/tasks/main.yml
@@ -11,7 +11,7 @@
     dest: /usr/local/bin/k3s
     owner: root
     group: root
-    mode: 755
+    mode: 0755
   when: ansible_facts.architecture == "x86_64"
 
 - name: Download k3s binary arm64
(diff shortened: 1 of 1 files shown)


St0rmingBr4in:
LGTM

PR #25, opened 2020-05-18
Fixes #24: Add more clarity around inventory file creation.

### LINKED ISSUE #24: Make inventory creation documentation slightly better
In the included `ansible.cfg`, the proper inventory file is configured:

```
inventory  = ./hosts.ini
```

This assumes the user has this repository, runs `ansible*` commands from the root directory, and has created a `hosts.ini` file in the base directory (alongside the `site.yml` playbook).

The README currently states:

> Add the system information gathered above into a file called hosts.ini. For example:

I think this would be more clear if it were something like "Add the system information gathered above into a file called `hosts.ini` in the same directory as this README file. There is a template in the `inventory` directory."

Additionally, because the path to the file is defined in `ansible.cfg`, it need not be specified when you run the playbook, so the playbook command could be, simply:

```
ansible-playbook site.yml
```

(Unless I'm reading the configuration wrong.)

Finally, if a `.gitignore` file is added to the repository with the `hosts.ini` file excluded, a user like me could clone the repository, create my custom `hosts.ini` file, and pull changes without fear of any conflicts or accidentally adding my local customized `hosts.ini` file to the repository.

geerlingguy:
Closes #24.﻿

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+hosts.ini
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ Master and nodes must have passwordless SSH access
 
 ## Usage
 
-Add the system information gathered above into a file called hosts.ini. For example:
+Add the system information gathered above into a file called `hosts.ini` in the same directory as this README file. There is a template in the `inventory` directory. For example:
 
 ```bash
 [master]
@@ -35,13 +35,12 @@ Add the system information gathered above into a file called hosts.ini. For exam
 [k3s_cluster:children]
 master
 node
-
 ```
 
 Start provisioning of the cluster using the following command:
 
 ```bash
-ansible-playbook site.yml -i inventory/hosts.ini
+ansible-playbook site.yml
 ```
 
 ## Kubeconfig
@@ -51,4 +50,3 @@ To get access to your **Kubernetes** cluster just
 ```bash
 scp debian@master_ip:~/.kube/config ~/.kube/config
 ```
-


St0rmingBr4in:
The host file is in the inventory directory so that the k3s-ansible repo can be added as a submodule of another git repository. This is how kubespray is handling this, it's much cleaner than having a .gitignore file in my opinion. That way you don't need to edit anything in this repository to make it work for your needs, you can version your configuration in the git repository that has k3s-ansible as a submodule. I think we should rather update the documentation to reflect the fact that you need to specify the inventory on which you want to run the playbook. (There could be multiple inventories) And remove the inventory configuration in ansible.cfg

St0rmingBr4in:
For instance, this is how I configure ansible with k3s-ansible:

```
.
├── .git
├── inventory (Variables specific to my cluster versionned)
│   ├── group_vars
│   └── hosts.ini
├── k3s-ansible (submodule)
│   ├── ansible.cfg
│   ├── group_vars
│   ├── hosts.ini
│   ├── README.md
│   ├── roles
│   └── site.yml
├── Makefile
├── post-cluster.yml (other playbook)
├── prepare-cluster.yml (other playbook)
├── README.md
└── Vagrantfile
```

geerlingguy:
I'll have to think on this a bit; for more ephemoral / test clusters, I like the 'clone/download repo, modify, and run it' technique that this PR _can_ enable. But it may not make for the best of both worlds.

Also, if you have `group_vars` inside an `inventory` directory, the group vars are not picked up if you use a `hosts.ini` anywhere besides that same directory (e.g. in the root directory). I usually see `group_vars` folders directly in the main folder, so I'm not sure why it's in a subdirectory in this case?

St0rmingBr4in:
I think it would be clearer if the default values for host.ini and `group_vars` were in a sample directory inside the inventory directory. 

That way it would enable the 'clone/download repo, modify, and run it' technique by just copying that directory and changing the variables as needed.

The quick start would look something like this:
https://github.com/kubernetes-sigs/kubespray/blob/master/README.md#usage

It would also be compatible if you want to use `k3s-ansible` as a submodule.
If the `group_vars` are defined in the root directory, it would not be possible to overwrite them without modifying the repository which less than ideal in the submodule technique.

PR #29, opened 2020-05-19
iptables-legacy for Raspian Buster

### LINKED ISSUE #28: Raspbian Buster - iptables
As pointed out in the official docs, Raspbian Buster needs to change to iptables-legacy for k3s:
https://rancher.com/docs/k3s/latest/en/advanced/#enabling-legacy-iptables-on-raspbian-buster

Should we add tasks for this?

rockaut:
Fixes #28 - Changing to iptables-legacy and ip6tables-legacy only for Buster

### DIFF
--- a/roles/raspbian/tasks/main.yml
+++ b/roles/raspbian/tasks/main.yml
@@ -20,7 +20,7 @@
   iptables:
     flush: true
   when:
-    - ansible_facts.distribution_release is search("buster")
+    - ansible_facts.lsb.description is match("[Rr]aspbian.*[Bb]uster")
 
 - name: Changing to iptables-legacy for Raspbian Buster
   alternatives:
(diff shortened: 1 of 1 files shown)


rockaut:
Sorry, I don't find why lint fails - it runs locally for me without msgs.

roles/raspbian/tasks/main.yml
```diff
@@ -16,8 +16,32 @@
     - ansible_facts.architecture is search("arm")
   register: boot_cmdline
 
+- name: Flush iptables before changing iptables-legacy
+  iptables:
+    flush: true
+  when:
+    - ansible_facts.distribution_release is search("buster")
```
St0rmingBr4in, line 23: Does this needs to run on Debian as well or just Raspbian?

rockaut: Dokumentation tells only for raspbian. Is Debian also called Buster?

rockaut: Thought this role would be loaded only on raspbian. Didn't realize you can load it anywhere.
Might be another catch. Will implement an additional when-rule

rockaut: Changed when to `- ansible_facts.lsb.description is match("[Rr]aspbian.*[Bb]uster")`.

PR #41, opened 2020-06-06
Have a sample inventory directory as a start point

zaherg:
Every time you edited the `hosts.ini` file it will create a conflict with the git repository, and since the `hosts.ini` file is ignored using `.gitignore` it is better to remove it from the cache and have an example file that people can use.

This way, we can just pull the changes without the need to revert our modification.


St0rmingBr4in:
Well, the best way would be to create multiple inventory directories since we have the same problem with `group_vars`. For example having by default : `inventory/sample/{hosts.ini,group_vars}`. Anyone cloning this repo could then create `inventory/myowncluster/{hosts.ini,group_vars}`. That way it would be possible to version the changes made to the created inventory directory using git and without interfering with the default one. Do you think you can do that?

zaherg:
give me a few min to check and see what I can do here.

zaherg:
@St0rmingBr4in can you have a look

README.md
```diff
@@ -23,7 +23,13 @@ Master and nodes must have passwordless SSH access
...
 ## Usage
 
-Add the system information gathered above into a file called `hosts.ini` in the same directory as this README file. There is a template in the `inventory` directory. For example:
+First create a new dirctory based on the `sample` directory within the `inventory` directory:
+
+```bash
+cp -R inventory/sample inventory/my-cluster
+```
+
+Second, edit `inventory/my-cluster/hosts.ini` to mactch the system information gathered above. For example:
```
St0rmingBr4in, line 32: typo "match"

PR #42, opened 2020-06-06
Fix Raspbian role to run correctly on Python 3.x (e.g. Ubuntu 20.04) Ansible host

fnord123:
Fix https://github.com/rancher/k3s-ansible/issues/40 .  New version works on both Python 2.x hosts (e.g. Debian 10, which it was previously working on) as well as Python 3.x hosts (e.g. Ubuntu 20.04, which it was broken on prior to this PR).


edenreich:
this pull request works on Buster Lite (32bits), I've just tested it

fnord123:
Thanks @z3r0l1nk and @geerlingguy. Do you need anything else from me prior to merging this? It should fix at least three different issues - #44, #49, and #54.

geerlingguy:
Nope!

St0rmingBr4in:
LGTM

Issue #13, opened 2020-05-12
Add CI to test the playbook


St0rmingBr4in:
A CI to test that the playbook is running successfully would be nice.

- Should we use Github action, Travis, Circle CI or other ?
- We need to test that systemd services works so running the playbook in docker might not be suitable

geerlingguy:
I typically use GitHub Actions or Travis CI, so I'm most familiar with those.

For my Ansible playbook testing (which usually requires testing systemd services), I use Molecule and some Docker images that have systemd built-in; for example: https://github.com/geerlingguy/ansible-for-devops/tree/master/molecule

St0rmingBr4in:
Molecule seems the way to go indeed.

fnord123:
One finding from https://github.com/rancher/k3s-ansible/issues/40 is that not only does the OS of the target machines (that will run k3s) matter, but that the OS of the Ansible host matters, which is obvious in retrospect.  So a full featured CI environment should include not only the set of OS that are supported as target devices, but also the set of OS that are supported as Ansible host devices.

PR #52, opened 2020-06-20
WIP: add tests

edenreich:
This pull request is still ongoing and related to #13 

There is probably a better way to structure molecule tests (I'm a bit new to molecule)
One thing I've noticed is that idempotence test is failing because we're deleting the downloaded binary everytime.

Feel free to add on this or restructure it :)


St0rmingBr4in:
Nice, definitely a step in the right direction

edenreich:
I've found that we're missing selinux-policy and libselinux-python yum packages on centos, but maybe it's just the docker image so i'm not entirely sure

edenreich:
Struggling to include roles from a different directory on a molecule test. any ideas ?
I'm trying to include k3s/master and k3s/node in k3s role.

I've tried **import_tasks**, **import_playbooks**, but then it yells at me that it cannot find jinja template :S

Wish it was simple as just given it a path attribute (unfortunately it's not the case):

    - name: Include master
      include_role:
        name: master
        path: master

St0rmingBr4in:
I will look into this, I think we need to get this working. It will catch a lot of little mistakes that inadvertently get merged.

FinalDes:
I also building my own molecule test have with follow coverage playbook 
```
---
- name: Converge
  vars:
    ansible_user: ansible
  ansible.builtin.import_playbook: ./../../site.yml
```
have follow error when run the molecule test
```
 UNREACHABLE! => {"changed": false, "msg": "Failed to create temporary directory.In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg to a path rooted in \"/tmp\", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p \"` echo ~/.ansible/tmp `\"&& mkdir \"` echo ~/.ansible/tmp/ansible-tmp-1614678369.911774-14191-28401085203069 `\" && echo ansible-tmp-1614678369.911774-14191-28401085203069=\"` echo ~/.ansible/tmp/ansible-tmp-1614678369.911774-14191-28401085203069 `\" ), exited with result 1", "unreachable": true}
```

dereknola:
Closing as out of date. Leaving the Issue open for a new push to add CI.

Issue #49, opened 2020-06-20
Activating cgroup support error


Nathan-j-hardee:
fatal: [192.168.1.150]: FAILED! => {"msg": "The conditional check 'raspbian is true' failed. The error was: template error while templating string: no test named 'true'. String: {% if raspbian is true %} True {% else %} False {% endif %}\n\nThe error appears to be in '/home/james/Downloads/k3s-ansible-master/roles/raspbian/tasks/main.yml': line 10, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Activating cgroup support\n  ^ here\n"}

ansible playbook fails to detect my OS version. 
cat /etc/osrelease:
NAME="Ubuntu"
VERSION="18.04.4 LTS (Bionic Beaver)"

this is running on an ARM64 system

fnord123:
PR https://github.com/rancher/k3s-ansible/pull/42 fixes this I believe.  Please try that.  Awaiting the PR to be accepted.

Nathan-j-hardee:
Understood, i tested #42 and it fixed the issue with the OS, it did introduce this error however:

fatal: [192.168.1.150]: FAILED! => {"changed": false, "msg": "Destination /boot/firmware/cmdline.txt does not exist !", "rc": 257}

fnord123:
You are correct, that is also reported in issue https://github.com/rancher/k3s-ansible/issues/40.    I'll start a new PR to address this.

vincentdyleung:
Hi, I believe PR #63 introduced a regression and broke this again.

Issue #40, opened 2020-06-06
Activate cgroup support failed


Vexet83:
When running the playbook after following the Readme and changing the user I get the following error when running the playbook.

TASK [raspbian : Activating cgroup support] **************************************************************************************************************************************************************************************************************************************************************************************************************************************************** 
fatal: [192.168.50.51]: FAILED! => {"msg": "The conditional check 'raspbian is true' failed. The error was: template error while templating string: no test named 'true'. String: {% if raspbian is true %} True {% else %} False {% endif %}\n\nThe error appears to be in '/home/brett/k3s-ansible/roles/raspbian/tasks/main.yml': line 10, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Activating cgroup support\n  ^ here\n"}
fatal: [192.168.50.215]: FAILED! => {"msg": "The conditional check 'raspbian is true' failed. The error was: template error while templating string: no test named 'true'. String: {% if raspbian is true %} True {% else %} False {% endif %}\n\nThe error appears to be in '/home/brett/k3s-ansible/roles/raspbian/tasks/main.yml': line 10, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Activating cgroup support\n  ^ here\n"}
fatal: [192.168.50.196]: FAILED! => {"msg": "The conditional check 'raspbian is true' failed. The error was: template error while templating string: no test named 'true'. String: {% if raspbian is true %} True {% else %} False {% endif %}\n\nThe error appears to be in '/home/brett/k3s-ansible/roles/raspbian/tasks/main.yml': line 10, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Activating cgroup support\n  ^ here\n"}
fatal: [192.168.50.204]: FAILED! => {"msg": "The conditional check 'raspbian is true' failed. The error was: template error while templating string: no test named 'true'. String: {% if raspbian is true %} True {% else %} False {% endif %}\n\nThe error appears to be in '/home/brett/k3s-ansible/roles/raspbian/tasks/main.yml': line 10, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Activating cgroup support\n  ^ here\n"}

[...]

PR #54, opened 2020-06-21
Fix cgroups on ubuntu x86 (Issue 40)

fnord123:
Fix for issues #40 and #49.  In both issues the problem is that while Ubuntu on RPi seems to need cgroup enabled, it doesn't need enabling on x86_64, and furthermore, the method of enabling on RPi breaks on x86_64.

Note that this also incorporates PR https://github.com/rancher/k3s-ansible/pull/42, so review and accept that PR request first for smaller PRs.


St0rmingBr4in:
Ok, looks good, we may need to test on a non raspberrypi based board on ARM to see if we encounter some problems. At least the playbook will not be broken on ubuntu on x86 which is the most common I think.

PR #59, opened 2020-07-14
Added creation instruction to avoid error on Ubuntu 20.04

dtourde:
Fixes : #60


fnord123:
This is an incorrect fix on Ubuntu x86 as Ubuntu x86 does not need this file created nor does it use it.  Pull request #54 is the correct fix as it only executes on Ubuntu ARM, which is the only place this file is needed/modified.

dtourde:
> This is an incorrect fix on Ubuntu x86 as Ubuntu x86 does not need this file created nor does it use it. Pull request #54 is a more correct fix as it changes things so that Ansible doesn't attempt to modify (or create) the file on Ubuntu x86, only on Ubuntu ARM, which is the only place this file is needed/modified.

After reviewing, PR #54 is indeed a better fix, close this one.

PR #63, opened 2020-07-19
Fixed Raspbian detection on 64-bit kernel, made reboot only do so if needed

JohnTheNerd:
Tested on a cluster of Raspberry Pi 3 and Raspberry Pi 4, all running Raspbian with a 64-bit kernel (`arm_64bit=1` in /boot/config.txt)

### DIFF
--- a/roles/raspbian/tasks/main.yml
+++ b/roles/raspbian/tasks/main.yml
@@ -5,7 +5,8 @@
       ( ansible_facts.architecture is search("arm") and
         ansible_facts.lsb.description is match("[Rr]aspbian.*[Bb]uster") ) or
       ( ansible_facts.architecture is search("aarch64") and
-        ansible_facts.lsb.description is match("Debian.*buster") ) %}True{% else %}False{% endif %}'
+        ansible_facts.lsb.description is match("Debian.*buster") or
+        ansible_facts.lsb.description is match("[Rr]aspbian.*[Bb]uster") ) %}true{% else %}false{% endif %}'
 
 - name: Activating cgroup support
   lineinfile:
@@ -13,7 +14,9 @@
     regexp: '^((?!.*\bcgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory\b).*)$'
     line: '\1 cgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory'
     backrefs: true
-  when: raspbian
+  register: cgroup
+  when:
+    - raspbian is true
 
 - name: Flush iptables before changing to iptables-legacy
   iptables:
@@ -36,4 +39,7 @@
 
 - name: Rebooting
   reboot:
-  when: raspbian
+  when:
+    - cgroup.changed
+  tags:
+    - skip_ansible_lint
--- a/roles/raspbian/tasks/main.yml
+++ b/roles/raspbian/tasks/main.yml
@@ -14,7 +14,7 @@
     regexp: '^((?!.*\bcgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory\b).*)$'
     line: '\1 cgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory'
     backrefs: true
-  register: cgroup
+  notify: reboot
   when:
     - raspbian is true
 
@@ -22,6 +22,7 @@
   iptables:
     flush: true
   when: raspbian
+  changed_when: false   # iptables flush always returns changed
 
 - name: Changing to iptables-legacy
   alternatives:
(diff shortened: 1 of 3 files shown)


roles/raspbian/tasks/main.yml
```diff
@@ -41,4 +43,6 @@
 - name: Rebooting
   reboot:
   when:
-    - raspbian is true
+    - cgroup.changed or ip4_legacy.changed or ip6_legacy.changed
+  tags:
+    - skip_ansible_lint
```
St0rmingBr4in, line 48: Is there a blocker to change this task to a handler?

JohnTheNerd: As far as I'm aware (and I might be wrong here), handlers apply to the entire group of tasks rather than a select few. The iptables flush always returns changed, so it would just always reboot anyway.

fnord123: Doesn't an IPTables flush take effect immediately? If that's the case, it shouldn't require a reboot.

A change to the boot command line does require a reboot to take effect, so it necessitates one.

It seems to me this change doesn't help anything.  The reboot already happens on Raspbian (where cgroups is changed) and the reboot shouldn't be done on other OS where only IPTables is changed.  What am I missing?

JohnTheNerd:
Accidentally clicked resolve, sorry about that

roles/raspbian/tasks/main.yml
JohnTheNerd: We are actually arguing the same point.

A change to the boot command certainly needs a reboot to take effect, but in its current state we reboot every single Raspbian machine every single time. Even if cgroups are already enabled. This change makes it so that we only reboot if cgroups are modified (seems that I accidentally put in the alternatives command too, I'll fix that) and not every single time the playbook is ran.

The IPTables command doesn't need a reboot, but always returns "changed". As it returns changed, (as far as I'm aware) any handler I put in the playbook will run regardless. This way, if I don't use a handler, we only reboot when it is needed and don't cause unnecessary downtime.

St0rmingBr4in: In that case we need to fix the changed status with `changed_when`. We would then report when something actually changed, and it will be possible to use a handler.

Also I don't think we actually need to reboot when `ip4_legacy.changed or ip6_legacy.changed` but let's keep it that way for now. I'll create an issue for this to fix in another PR.

JohnTheNerd: I have indeed noticed the `ip4_legacy.changed` and `ip6_legacy.changed` and have fixed them in the merge commit. They do not need a reboot to take effect. I didn't notice that I did it as part of a merge commit, I can fix that if you'd like.

St0rmingBr4in: if you already tested it, let's kill two birds with one stone.
Just update the commit/pr description to state the change in reboot strategy. 

If cgroups.changed is working as expected, let's create a handler it'll be cleaner. :)

JohnTheNerd: Sounds good! It's all tested with a hybrid cluster of Raspberry Pi 3 and 4's as part of a larger playbook (I run the roles separately and don't use `site.yml` itself, but that shouldn't matter).

I will squash all commits to one, create the handler (after adding `changed_when: false` to IPTables, as the task always returns changed otherwise), test the playbook and request a re-review.

site.yml
```diff
@@ -8,6 +8,9 @@
     - role: download
     - role: raspbian
     - role: ubuntu
+  handlers:
+    - name: reboot
```
St0rmingBr4in, line 12: We only need it for the raspbian role, why not putting it in the `roles/raspbian/handlers` folder?

St0rmingBr4in:
Alright I finally took the time to test the change myself. LGTM thank you for your contribution :)

roles/raspbian/tasks/main.yml
```diff
@@ -5,20 +5,24 @@
...
 - name: Activating cgroup support
   lineinfile:
     path: /boot/cmdline.txt
     regexp: '^((?!.*\bcgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory\b).*)$'
     line: '\1 cgroup_enable=cpuset cgroup_memory=1 cgroup_enable=memory'
     backrefs: true
-  when: raspbian
+  notify: reboot
+  when:
+    - raspbian is true
```
vincentdyleung, line 19: Hi, I think this should be raspbian rather than raspbian is true? See PR #42.

JohnTheNerd: that's true, that PR wasn't merged by the time I made this which is why I didn't notice it. although, this PR has already been merged and I already cleaned up the forked repo, so it's probably best to fix in a separate PR

vincentdyleung: Thanks @JohnTheNerd I will create a PR and fix it.

vincentdyleung: I have raised PR #74.

PR #65, opened 2020-07-21
Pass through extra arguments to the nodes with extra_node_args group …

b-m-f:
…variable


St0rmingBr4in:
Maybe it would be more consistent if it was named extra_agent_args. What do you think? Otherwise it looks good!

b-m-f:
Updated it!

St0rmingBr4in:
LGTM. Thank you for your contribution

PR #66, opened 2020-07-22
reset role updated to killall containerd daemons and umount k3s filesystems

stafwag:
killall containerd-shim-runc-v2 added
umount k3s filesystems added

### DIFF
--- a/roles/reset/tasks/main.yml
+++ b/roles/reset/tasks/main.yml
@@ -10,9 +10,10 @@
     - k3s-node
 
 - name: Killall containerd-shim-runc-v2
+  register: killall_containerd_shim_runc_v2
   command: killall -9 containerd-shim-runc-v2
+  changed_when: "killall_containerd_shim_runc_v2.rc == 0"
   failed_when: false
-  changed_when: false
 
 - name: Umount k3s filesystems
   include_tasks: umount_with_childeren.yml
--- a/roles/reset/tasks/main.yml
+++ b/roles/reset/tasks/main.yml
@@ -4,10 +4,25 @@
     name: "{{ item }}"
     state: stopped
     enabled: no
-  failed_when: false
+  failed_when: False
   with_items:
     - k3s
     - k3s-node
+
+- name: Killall containerd-shim-runc-v2
+  command: killall -9 containerd-shim-runc-v2
+  failed_when: False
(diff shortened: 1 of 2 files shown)


roles/reset/tasks/main.yml
```diff
@@ -4,11 +4,26 @@
...
-  failed_when: false
+  failed_when: False
   with_items:
     - k3s
     - k3s-node
 
+- name: Killall containerd-shim-runc-v2
+  command: killall -9 containerd-shim-runc-v2
+  failed_when: False
+  changed_when: False
```
St0rmingBr4in, line 15: Why should we set `changed_when` to false? Default value should be ok?

roles/reset/tasks/main.yml
```diff
@@ -4,11 +4,26 @@
     name: "{{ item }}"
     state: stopped
     enabled: no
-  failed_when: false
+  failed_when: False
```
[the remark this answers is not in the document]
stafwag: Ok. updated in commit 637e49d

roles/reset/tasks/main.yml
stafwag: I updated it to verify the exit code of killall. killall reports exitcode 0 is something is killed.
Process killed -> change
Nothing to kill -> no change

 updated in commit afc6c8a

St0rmingBr4in: Yep your change is good. We can't use the default value because killall outputs something on stderr and ansible interpret this as a change.

roles/reset/tasks/main.yml
```diff
@@ -9,6 +9,22 @@
     - k3s
     - k3s-node
 
+- name: Killall containerd-shim-runc-v2
+  register: killall_containerd_shim_runc_v2
+  command: killall -9 containerd-shim-runc-v2
```
St0rmingBr4in, line 14: Sorry for not mentioning in the first review. Maybe we should prefer using `pkill` instead of `killall`, not everyone has `psmisc` installed on their system.

stafwag: np, I updated & tested it to use pkill on the full command line.

updated in commit 7bab4cb

St0rmingBr4in:
LGTM Thank you for your contribution :)

Issue #56, opened 2020-07-04
"{{ items }}" in /prereq/tasks/main.yml should be "{{ item }}"?


guofengzh:
"{{ items }}" in [this line](https://github.com/rancher/k3s-ansible/blob/master/roles/prereq/tasks/main.yml#L35) should be "{{ item }}".

Ansible version: 2.9.10
OS: CentOS Linux release 7.8.2003 (Core)

After changed to {{ item }}, the task finished successfully.

evrardjp:
You're right. Did you propose a PR?

guofengzh:
PR? how could I do it?

evrardjp:
You could fork this repository (it's one click on the main repo page) to your own repository, change the line in your repository and commit it (either using github's web interface or using your favorite git client), and then there will be a link on your forked repository's github page to "Create a pull request" in this repository. Click on it, and it will automatically propose the change for the reviewers of this project.

Aditional tip: when writing the commit message, you can type: "Fixes: #56" and merging your PR will auto-close this issue!

PR #78, opened 2020-08-15
fix: typo in item variable

### LINKED ISSUE #77: [ansible-galaxy]Typo error causing error in Centos execution
I'm facing with this error when I'm using Ansible Galaxy version (in ansible-galaxy branch):

TASK [k3s-ansible/roles/prereq : Set bridge-nf-call-iptables (just to be sure)] ***
fatal: [10.0.10.11]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'items' is undefined\n\nThe error appears to be in '/roles/k3s-ansible/roles/prereq/tasks/main.yml': line 33, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Set bridge-nf-call-iptables (just to be sure)\n  ^ here\n"}

That variable should be "item":
https://github.com/rancher/k3s-ansible/blob/f91dfcfc8e2e94d6ff687c3d0ecc7805d38e8517/roles/prereq/tasks/main.yml#L35

renanqts:
It should fix #77

### DIFF
--- a/roles/prereq/tasks/main.yml
+++ b/roles/prereq/tasks/main.yml
@@ -32,7 +32,7 @@
 
 - name: Set bridge-nf-call-iptables (just to be sure)
   sysctl:
-    name: "{{ items }}"
+    name: "{{ item }}"
     value: "1"
     state: present
     reload: yes


dereknola (approved):

dereknola:
Resolved in https://github.com/k3s-io/k3s-ansible/issues/56

PR #85, opened 2020-08-30
update k3s version

Eagle-One1:
Hello, 

It's just a minor change but I did some work on a role I have made for k3s, https://github.com/Eagle-One1/ansible-k3s/tree/init
I think some changes I made could be integrates in your role, like the download, I nearly handle upgrades by checking if the checksum of the k3s binaries are different or not from the version in github, the only thing missing is to reboot when the k3s binary is changed.

If your interested in merging some parts I'll work on it and make another pull request


rdxmb:
can this be merged?

PR #79, opened 2020-08-16
Moved Raspberry Pi specifics settings to raspberrypi role

stafwag:
* Moved Raspberry Pi specifics settings to the raspberrypi role. Raspberry Pi detection is done by analyzing /proc/cpuinfo and /proc/device-tree/model.
* This adds support for ARM64 virtual machines and other ARM boards.
* Added support for multiple operating systems and versions, as I'm not sure cgroups needs to enable all GNU/Linux distributions on the Raspberry Pi.
* Added support for CentOS on the Raspberry Pi
* Closes issues #73 #62 #68 #49

PR #87, opened 2020-09-04
Fix raspbian role on centos

miff2000:
CentOS based OSes don't have the contents of `ansible_facts.lsb.description` populated by Ansible facts gathering, so the detection of Raspbian causes the playbook to fail with this on Ansible 2.8.11:

```
TASK [raspbian : Test for Raspbian] ******************************************************************************************************************************************
fatal: [k3s-node-01]: FAILED! => {}

MSG:

Unexpected templating type error occurred on ({% if ( ansible_facts.architecture is search("arm") and ansible_facts.lsb.description is match("[Rr]aspbian.*[Bb]uster") ) or (              ansible_facts.architecture is search("aarch64") and ansible_facts.lsb.description is match("Debian.*buster") or ansible_facts.lsb.description is match("[Rr]aspbian.*[Bb]uster             ") ) %}true{% else %}false{% endif %}): expected string or bytes-like object
```

on Ansible 2.9.13 this is the error:
```
TASK [raspbian : Test for Raspbian] *******************************************************************************************************************************************************
fatal: [k3s-node-01]: FAILED! => {}

MSG:

The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'description'

The error appears to be in '/home/matt.calvert/git/ansible/k3s/roles/raspbian/tasks/main.yml': line 2, column 3, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

---
- name: Test for Raspbian
  ^ here
```

This MR changes the logic so that `raspbian` is set to `false` by default, and it gets set to `true` by the existing `set_fact` task if a raspbian OS is detected.

### DIFF
--- a/roles/raspbian/defaults/main.yml
+++ b/roles/raspbian/defaults/main.yml
@@ -1,0 +1,3 @@
+---
+
+raspbian: false
--- a/roles/raspbian/tasks/main.yml
+++ b/roles/raspbian/tasks/main.yml
@@ -1,12 +1,18 @@
 ---
 - name: Test for Raspbian
   set_fact:
-    raspbian: '{% if
-      ( ansible_facts.architecture is search("arm") and
-        ansible_facts.lsb.description is match("[Rr]aspbian.*[Bb]uster") ) or
-      ( ansible_facts.architecture is search("aarch64") and
+    raspbian: true
+  when: >-
+    (
+      ansible_facts.architecture is search("arm") and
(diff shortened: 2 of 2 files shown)


miff2000:
Those linting issues are not my doing. I can fix though if you want?

dereknola:
Overridden by https://github.com/k3s-io/k3s-ansible/pull/79

PR #92, opened 2020-09-20
Use checksum with get_url

### LINKED ISSUE #67: validate k3s checksum on download roles
validate k3s checksum on download roles and shouldn't be deleted if checksum match when k3s already exists

lentzi90:
When the checksum is used there is no need to delete the binary first, 
as it will be compared to the checksum and downloaded and replaced only 
as needed.

Fixes #67


St0rmingBr4in (approved):
LGTM, thank you

PR #95, opened 2020-10-07
Added external hdd

xussof:
aded a role to format, mount and share through nfs a hard drive connected to master node


St0rmingBr4in:
Thank you for your work, but formating and exposing a drive through nfs is not in the scope of this project. Your role can definitely be useful, but this does not belong in the k3s-ansible repo.

PR #107, opened 2020-12-14
Add Alpine Linux configuration

### LINKED ISSUE #69: Please add ansible support for openrc-based systems
I'm trying to get this running on Alpine.

jiayihu:
This adds support for Alpine Linux. There is a tricky question about how to best handle reboot because it depends on how one is using Alpine. If it is diskless as many do, then reboot cannot be done before committing with `lbu`. 

The approach I chose is to just notify the developer. This allows to manually check if the installation is correct before committing and rebooting. After reboot, the service starts and k3s is already working correctly without a second run of the playbook.

This should close #69 , or at least provide a good example of how to do it.

### DIFF
--- a/roles/k3s/master/tasks/main.yml
+++ b/roles/k3s/master/tasks/main.yml
@@ -1,20 +1,10 @@
 ---
 
-- name: Copy K3s service file
-  register: k3s_service
-  template:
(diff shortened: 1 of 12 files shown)


itwars (approved):
Just work fine

PR #113, opened 2021-01-12
Shutdown restart nodes

ElioLopez:
Adding useful operations playbooks for reboot and safely shutdown the servers


St0rmingBr4in:
I don't think we actually need playbooks for those operations. I think using ad-hoc commands would be preferable. If you want you could add in the readme how to use ad-hoc commands to reboot or shutdown all the nodes. Will close this PR, feel free to reopen if you decide to add those commands to the readme.

PR #116, opened 2021-01-24
Reorganize the filesystem umount order.

stafwag:
* Clean /var/lib/kubelet before /var/lib/rancher/k3s.
* Umount the filesystem tree in reverse order, to ensure the root
  is umounted last.


St0rmingBr4in (approved):
LGTM

PR #117, opened 2021-01-28
Revert "Add Alpine Linux configuration"

itwars:
Reverts k3s-io/k3s-ansible#107


jakuboskera:
Why was this reverted?

timberhill:
Hey @itwars, Alpine support looked quite useful, why was it removed?

fsdrw08:
We need alpine support, please add it back

PR #123, opened 2021-04-20
Update README.md remote cluster access instructions

trevorackerman:
Provide details about how to update kubeconfig in a non-destructive manner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,45 @@ ansible-playbook site.yml -i inventory/my-cluster/hosts.ini
 
 ## Kubeconfig
 
-To get access to your **Kubernetes** cluster just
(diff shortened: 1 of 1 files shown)


St0rmingBr4in:
I think it would be better to keep the easy way as the default manner of doing it.

Of course ansible could fetch the kubeconfig file on the cluster and store it as an artifact on the host, that would avoid having to do manual steps like those.

Maybe switching around your explanation and the previous one and adding a warning to state that the scp will override any already present configs would be better I think.

trevorackerman:
@St0rmingBr4in I made the suggested change. Let me know what you think.

dereknola:
Closing as it makes the README overly complex.

PR #125, opened 2021-05-13
Abort the play on all hosts if any error happens in the k3s_cluster building process

mgsweet:
When I tried to run the playbook, the download process failed because of some network issues. Since only some of the nodes can download the requirements for k3s, the cluster setup tasks will fail. Instead of stopping the task manually, I think using `any_errors_fatal: true` in the k3s_cluster building process will be better.

### DIFF
--- a/site.yml
+++ b/site.yml
@@ -1,6 +1,7 @@
 ---
 
 - hosts: k3s_cluster
+  any_errors_fatal: true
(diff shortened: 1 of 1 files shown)


dereknola:
Closing as out-of-date, feel free to reopen and rebase.

Issue #32, opened 2020-05-20
Multi-master support for K3s Ansible playbook?


geerlingguy:
This is something I think we might be able to get configured in the Ansible playbook, but I didn't see (at a glance at least) if it was something supported by this playbook yet; namely, a multi-master configuration with an external database: [High Availability with an External DB](https://rancher.com/docs/k3s/latest/en/installation/ha/).

In this playbook's case, maybe it would delegate the task of configuring an external database cluster to the user (e.g. use a separate Ansible playbook that builds an RDS cluster in Amazon, or a separate two or three node DB cluster on some other bare metal servers alongside the K3s cluster), but then how could we make it so this playbook supports the multi-master configuration described in the docs page linked above.

sethcohn:
This role does this:
https://github.com/PyratLabs/ansible-role-k3s

MrDienns:
Would love to have this supported by Rancher out of the box. Any plans on implementing this feature in this repository?

2fst4u:
Now that internal HA is implemented, even if only experimental, does this playbook automatically install and set up a cluster to utilise this feature already or if not, can it be updated to support this?

St0rmingBr4in:
I've implemented the base mechanism in PR #97 It setups the HA with the embedded etcd database mode.
It is still missing some parts and still need some work to be mergable. But if you want to play a bit you could try to checkout the branch and tell me how it goes :)

2fst4u:
I'd personally love to but my pitiful single node cluster is clearly not going to be much help!

I'd love to know if someone else can confirm it though. This would be so awesome.

St0rmingBr4in:
Maybe it's time to create a proper vagrant file to test it, or finish the work in #52 to test it inside docker.

clarkezone:
Would love this as well

2fst4u:
> I've implemented the base mechanism in PR #97 It setups the HA with the embedded etcd database mode.
> It is still missing some parts and still need some work to be mergable. But if you want to play a bit you could try to checkout the branch and tell me how it goes :)

How is this going? I see the PR looks pretty ready to go. Did it work nicely in testing?

St0rmingBr4in:
> > I've implemented the base mechanism in PR #97 It setups the HA with the embedded etcd database mode.
> > It is still missing some parts and still need some work to be mergable. But if you want to play a bit you could try to checkout the branch and tell me how it goes :)
> 
[...]

PR #131, opened 2021-06-24
Add control plane api fail-over

mmansell83:
This PR does a few things but the primary item is it uses keepalived across the server/master nodes to fail-over an `apiserver_endpoint` virtual IP address. This addresses the need for an external load balancer as discussed in #32.
Using keepalived is configurable and the default behavior is disabled.

This PR also updates some of the role nomenclature to match that of the upstream Rancher documentation i.e. master -> server and node -> agent.

Finally, I have also migrated all of the short Ansible module names to their new naming convention per the Ansible 2.10 release.

As a note I'm new to submitting PRs on Github so if there is anything I am missing please let me know. Feedback is always welcome!


dereknola:
Closing this. With the newest changes, this PR is heavily offset and will need a major overhaul. Also, going forward, this repo will not be implementing any VIP by default. The `api_endpoint` has now been merged, users can easily setup a VIP themselves and then connect into afterwards.

PR #134, opened 2021-07-10
don't try to set ipv6 forward if ipv6 not enabled

clambin:
k3s-ansible playbook will fail if the node doesn't have ipv6 disabled. This PR fixes that.


allentc:
This fix prevents the sysctl task that sets net.ipv6.conf.all.forwarding=1 from running unless the ansible fact ansible_all_ipv6_addresses fact contains IPv6 addresses.  Likewise, the "Enable IPv4 forwarding" task should also have a "when: ansible_all_ipv4_addresses" check for IPv6 only hosts.

Issue #135, opened 2021-08-06
Only seeing master, No nodes


anthonybudd:
I ran this playbook on 3x Pi 3's. The install seems to be successful however when I run `kubectl get nodes`
 I only see the master node, there are no other worker nodes?

```sh
> ansible-playbook site.yml -i inventory/pi-cluster/hosts.ini

> scp pi@192.168.1.64:~/.kube/config ~/.kube/config

> kubectl get nodes
NAME          STATUS   ROLES    AGE   VERSION
raspberrypi   Ready    master    1h   v1.17.5+k3s1
```

Does anyone have any suggestions about how I can get the other 2 nodes talking to the master?

**Possibile Issue?**
When I run `nmap 192.168.1.0/24` I've noticed that the master node doesn't have port [6443](https://github.com/k3s-io/k3s-ansible/blob/master/roles/k3s/node/templates/k3s.service.j2#L10) open? Is this preventing the nodes for communicating to kuber on the master node?
```sh
> nmap 192.168.1.0/24

Nmap scan report for raspberrypi (192.168.1.64) 👈 Master Node
Host is up (0.0030s latency).
Not shown: 999 closed ports
PORT   STATE SERVICE
22/tcp open  ssh 👈 Should port 6443 be listed here to?

Nmap scan report for raspberrypi (192.168.1.90)
Host is up (0.0030s latency).
Not shown: 999 closed ports
PORT   STATE SERVICE
22/tcp open  ssh

Nmap scan report for raspberrypi (192.168.1.91)
Host is up (0.0012s latency).
Not shown: 999 closed ports
PORT   STATE SERVICE
22/tcp open  ssh
```

JasonCarter80:
Same behavior here

JasonCarter80:
If I use v1.20.9+k3s1 it works fine, issues seems to be with latest release.

JasonCarter80:
With v1.21.3+k3s1 it seems to get stuck on [k3s/node : Enable and check K3s service], I left it running over night, no change.   
Regardless, master only no nodes.

This morning, I reset the whole thing, updated to last version: v1.20.9+k3s1, which completed the process but still, no nodes.

JasonCarter80:
My problem is with duplicate hostnames, I imaged 4 SDCARDS for my RasPi cluster all with the same name.   Added a hostname task to rename which resolved my issues here.

anthonybudd:
**SOLVED**
[...]

PR #136, opened 2021-08-07
Random node-id

anthonybudd:
This solves the issue of devices not connecting to the cluster if they have the same hostname as another node on the cluster.

See: https://github.com/k3s-io/k3s-ansible/issues/135

@reviewer, should `--with-node-id node-$(openssl rand -base64 12)` be in [all.yml](https://github.com/k3s-io/k3s-ansible/blob/master/inventory/sample/group_vars/all.yml#L7) or [roles/k3s/node/templates/k3s.service.j2](https://github.com/k3s-io/k3s-ansible/blob/master/roles/k3s/node/templates/k3s.service.j2#L10). All.yml is easier for users to remove if it's not needed, k3s.service.j looks cleaner.


dereknola:
The sample vars will remain empty. It will be up to the individual user to enforce specific requirements for their setup.

PR #143, opened 2021-09-08
use full path for k3s binary

mhahl:
Don't rely on `/usr/local/bin` being in the `PATH`.


MartijnBNG:
I could've used this one today 👍 
Although I used this instead:
```
    environment:
      PATH: /usr/local/bin:{{ ansible_env.PATH }}
```

mhahl:
@MartijnBNG great idea! I'll update the PR with that change.

dereknola (approved):

PR #144, opened 2021-09-14
Add HTTP_PROXY, HTTPS_PROXY and NO_PROXY envs to k3s service

danielfbm:
Adds proxy settings for k3s service

I would love to add some conditionals and avoiding settings this env vars if the values are not provided, but this is my first time using Ansible, any help will be much appreciated 😄


dereknola:
Closing in favor of https://github.com/k3s-io/k3s-ansible/pull/237 which handles the generic case.

PR #158, opened 2021-12-14
Fix omissions and errors in the reset role that leave artifacts after successful reset.yml playbook run

jon-stumpf:
The current **reset** role does not remove all network artifacts (#159).  I updated the **reset** role to perform the same actions in the same order as **k3s-uninstall.sh** as provided by https://get.k3s.io.  I used v1.22.4+k3s1 as the reference.

PR #164, opened 2021-12-19
Implemented many install flags as found in the install script from https://get.k3s.io

jon-stumpf:
Similar to my work in reset.yml (#158), I implemented many of the install flags that are available when using the install script from [https://get.k3s.io](https://get.k3s.io).  These are all found in **inventory/sample/group_vars/all.yml**.  All of the variables found in inventory/sample/group_vars/all.yml (except **ansible_user**) have moved to **playbook/group_vars/all.yml**; and, the two playbooks, **reset.yml** and **site.yml**, have moved into the new **playbook/** directory.

This work builds on my previous pull request (#158).


dereknola:
Closing this PR. Instead of adding individual arguments as global vars, this repo now supports a variety of ways to add arguments including:
- extra_server_args: added to the service exec command
- extra_agent_args: added to the service exec command
- extra_service_envs: environmental variables added to the service
- config_yaml: yaml values added as the k3s config.yaml before install. 
- extra_manifests: Helm Manifests on the management host that are transferred to targets and automatically installed by K3s.

PR #151, opened 2021-11-06
Fix Raspberry tasks for Debian

kcirrr:
Debian Buster was previously recognized as Raspbian, and there were no tasks for Debian Bullseye.
This pull requests changes the os detection tasks and adds tasks for Debian to make sure the behaviour for the previous Debian versions stays the same. It also adds Debian 11 specific tasks.

PR #169, opened 2022-01-25
Added bullseye to Raspbian fact conditional list

sdstolworthy:
Latest versions of Raspbian are based on Bullseye. I added "bullseye" to the conditional that sets the Raspbian fact.


cwayne18:
Thanks for the contribution! Can you please make sure to sign your commit (with git commit -s)

sdstolworthy:
There we go :)

kcirrr:
This would break cause the role to fail to run on regular Debian. The cmdline task would fail since it's on a different path.
PR #151 would fix the role for Raspbian and regular Debian.

PR #185, opened 2022-04-09
Add support raspbian bullseye

husjon:
This PR adds support for raspbian bullseye.
It has been tested on newly flashed pi3b+ nodes (combination of buster and bullseye) with only SSH enabled.
(I do currently not have pi4 nodes in my possession)

Test run output: https://gist.github.com/husjon/00965e50afffd8865e8ac3db66edb30f

What's been changed:
 * Added check for Raspbian Bullseye
 * Added task switch for Raspbian distribution codenames
 * Removed requirement for `iptables` on Rasbpian Bullseye (this is because bullseye comes with `nftables` preinstalled)
 * Removed 'cgroup_enable=cpuset' for bullseye.
   On bullseye using `Pi3B+` this seem to bring the server to its knees, seeing average loads up to 15 while agents are added.
   This makes it so that the agents are unable to communicate with the server. (screenshot in comments)


roles/raspberrypi/tasks/main.yml
```diff
@@ -33,18 +33,29 @@
     - raspberry_pi|default(false)
     - ansible_facts.lsb.description|default("") is match("Debian.*buster")
 
+- name: Set detected_distribution to Raspbian (ARM64 on Debian Bullseye)
+  set_fact:
+    detected_distribution: Raspbian
+  when:
+    - ansible_facts.architecture is search("aarch64")
+    - raspberry_pi|default(false)
+    - ansible_facts.lsb.description|default("") is match("Debian.*bullseye")
```
husjon, line 42: I made the detection task a separate task for bullseye to be explicit for bullseye, but if prefered, this can be combined with the buster match.
A few comments on this would be appreciated.

Example change:
```suggestion
    - ansible_facts.lsb.description|default("") is match("Debian.*(buster|bullseye)")
```

husjon:
Here's a screenshot of the load from the server while adding agents.
I was lucky to have the SSH connection open during this, the connection is barely working and `htop` is updating only every now and then.

image

JohanNicander:
Nice work! I would have solved it in nearly the same way. However, adding the codename (which is a one to one map of major version) is redundant imo, and I would suggest we use only one of major version and codename.

husjon:
Thanks for the feedback, I don't mind either solution, I just found it easier to read the codenames rather than version numbers for the task filenames. :)
I do however realize that `major_version` is used in the prior task setting the facts [raspberrypi/tasks/main.yml#L36-L40](https://github.com/k3s-io/k3s-ansible/blob/master/roles/raspberrypi/tasks/main.yml#L36-L40), so might as well continue this trend.

I'll prepare a commit with the necessary changes.

ozgurgul (approved):
Suggested change works!.

ozgurgul:
Suggested change works!.

kraihn:
@ozgurgul and @JohanNicander, is there a reason this fix is approved but hasn't been merged?

KI7ODK:
> @ozgurgul and @JohanNicander, is there a reason this fix is approved but hasn't been merged?

I had to implement this fix the other day setting up a new Raspberry Pi cluster with Raspberry Pi OS, so I'm wondering the same thing. This seems like a simple fix that should be merged with little issue.

ozgurgul:
> > @ozgurgul and @JohanNicander, is there a reason this fix is approved but hasn't been merged?
> 
> 
> 
> I had to implement this fix the other day setting up a new Raspberry Pi cluster with Raspberry Pi OS, so I'm wondering the same thing. This seems like a simple fix that should be merged with little issue.

I'm not a contributor to this repo. Hence I can't merge this PR but otherwise I share the same thoughts.

husjon:
Hi @St0rmingBr4in what is required for this PR to be merged?
Thanks in advance

berkobob:
Please can this PR get merged sooner rather than later. Thanks.

akrami:
tested this on my raspberry pi cm4 cluster and totally works 👍

nullset2:
I'm SO tempted to use this. I tried to spin up my cluster with an `aarch64` OS instead of `armv7` and it didn't really work out for me, but it seems that this would solve my issue completely.

I say this PR should be merged!

dereknola:
Closing in favor of changes in https://github.com/k3s-io/k3s-ansible/pull/151

PR #187, opened 2022-04-24
Add --with-node-id to prevent colliding names

JoePassanante:
Noticed that, when running on a cluster of raspberry pis running Ubuntu 20 LTS, all 3 nodes would attempt to use "ubuntu" as their name resulting in the cluster failing to start. 

Fixed by adding --with-node-id to the exec start command.


dereknola:
Adding this as a default is not warranted, you are free to modify you own version of inventory with `extra_server_args=--with-node-id`.  By default, K3s assumes that  each node will have a unique hostname.

Issue #188, opened 2022-04-27
parsing error when running  playbook Ubuntu 20.04 Server on Rasberry pi 4 cluster. 


fdsig:
Playbook runs ⏯ great. I have been having an issue with parsing the master node token. 
 Playbook returns a parsing error for the master node token from line 40 on main.yml in the roles/k3s/master/tasks directory.
```
├── roles
│   ├── k3s
│   │   ├── master
│   │   │   ├── tasks
│   │   │   │   └── main.yml
```
```yml
  token: "{{ node_token.content | b64decode | regex_replace('\n', '') }}"
```

While elsewhere errors returned at end of play do not always correctly respond to the underlying issue in my experience this appears to have been fixable by adding ```| trim```  to remove trailing leading white space as in : 

```yml
 token: "{{ node_token.content | b64decode | regex_replace('\n', '') | trim }}"
```

jinja documentation on this can be found here:

[https://github.com/lxhunter/ansible-filter-plugins/blob/master/docs/jinja_functions.md#trim](https://github.com/lxhunter/ansible-filter-plugins/blob/master/docs/jinja_functions.md#trim)

dereknola:
Closing as the ansible playbook no longer supports dynamically pulling the token from inital server. A token must now be provided. This line has been removed.

PR #189, opened 2022-04-27
adding | trim to remove whitespace

fdsig:
fixes issue #188 that I logged. simply adds trim in in line with jinja and ansible docs.


dereknola:
Closing as the ansible playbook no longer supports dynamically pulling the token from inital server. A `token` must now be provided.

PR #191, opened 2022-05-16
Make apiserver port user-configurable

bcbrookman:
Added a new `apiserver_port` variable to allow users to specify an alternate port for the loadbalanced apiserver endpoint. This is required if port 6443 is already in use on the loadbalancer (e.g., if you have multiple clusters behind the same loadbalancer frontend IP.) 6443 is set as a default to maintain compatibility in case `apiserver_port` is not defined.

Note: This only configures the port for the loadbalanced apiserver endpoint; nodes themselves will still listen on 6443.


St0rmingBr4in (approved):

bcbrookman:
Added an additional commit to add FQCNs for builtin actions and pass the lint check.

dereknola (approved):

PR #194, opened 2022-06-25
According to https://github.com/k3s-io/k3s/issues/2435

hanhsuan:
playbook should to remove useless netns while reset.


dereknola:
Resolved with https://github.com/k3s-io/k3s-ansible/pull/230, which now uses the K3s-Killall.sh/uninstall.sh scripts

PR #199, opened 2022-07-26
Add support for SELinux

tylergmuir:
Adds support for SELinux on CentOS 7 and 8 as well as RedHat Enterprise Linux 7 and 8.

### DIFF
--- a/inventory/sample/group_vars/all.yml
+++ b/inventory/sample/group_vars/all.yml
@@ -7,4 +7,4 @@ extra_server_args: ""
 extra_agent_args: ""
 selinux_mode: disabled
 selinux_policy_version: v1.1.stable.1
-selinux_rpm_version: "{{ selinux_policy_version[1:] | regex_replace('\.[a-zA-Z]*\.', '-') }}"
+selinux_rpm_version: "{{ selinux_policy_version[1:] | regex_replace('\\.[a-zA-Z]*\\.', '-') }}"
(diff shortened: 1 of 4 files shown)


tylergmuir:
#199

dereknola:
SELinux support is now automatically handled by using the K3s install script, added in https://github.com/k3s-io/k3s-ansible/pull/228

tylergmuir:
@dereknola The install script does handle the installation of the selinux rpm now, but in the prereqs task in this repo, the first step is to disable SELinux. In addition to that, it would probably be good to document that `--selinux` would need to be added to the `extra_server_args` and `extra_agent_args` for it to work appropriately.

Also, if you look at the changed files for this PR, it also fixes some issues that someone running a RHEL distro would run into, such as firewalld being enabled (which conflicts with k3s) and that the distro name in the `when` statements is incorrect for modern versions of ansible.

I would be happy to update my fork or open an updated PR if you would like, but I feel there are still issues here that should be addressed.

dereknola:
Hey @tylergmuir Those issues you mentioned are valid. I have an open issue for https://github.com/k3s-io/k3s-ansible/issues/234. I will open another issue to track SELinux working properly and make sure that it works correctly. As for the RHEL distro when statements, https://github.com/k3s-io/k3s-ansible/commit/45289ba7d9ffd6e99dea1cfd7e5c63f266a599be merged today and should handle them correctly now. 

Lookout for stuff tomorrow and next week.

Issue #160, opened 2021-12-14
The reset role does not use k3s_server_location


jon-stumpf:
**k3s_server_location** was introduced in #133 but is only used during the installation (**site.yml** playbook).  The same location should also be used during uninstallation (**reset.yml** playbook).

dereknola:
Resolved with https://github.com/k3s-io/k3s-ansible/commit/cd725780f5a58bb8a51c53ed625501ab3e5af475

Issue #127, opened 2021-06-11
Add support for embedded DB HA mode


kopwei:
Since v1.19.5, K3s added support for HA mode with embedded ETCD and I hope this could be supported in Ansible scripts.

The document is described at [K3s HA Embedded mode](https://rancher.com/docs/k3s/latest/en/installation/ha-embedded/)

Generally. First we shall define "K3S_TOKEN" as an input value, the value should be a string **without special chars**. e.g. "mySecret123", 
- On first master node, k3s should start via command "K3S_TOKEN='mySeceret123' k3s server --cluster-init [--extra server config]"
- On other master, k3s should start via command "K3S_TOKEN='mySecret123' k3s server --server=https://<IP_OF_1st_MASTER>:6443 [--extra server config]"
- For the rest worker nodes, k3s shall start as k3s-agent as before

212850a:
It's already implemented in this repo - https://github.com/k3s-io/k3s-ansible/tree/k3s-ha ;)

2fst4u:
It's also in this issue https://github.com/k3s-io/k3s-ansible/issues/32

dtaivpp:
> will these changes be merged into master?

Curious about this as well. Seems to automatically detect if there are multiple masters so I don't see why it couldn't be added back in.

PR #97, opened 2020-10-28
Support HA mode with embedded DB

St0rmingBr4in:
This enables initializing a cluster in HA mode with an embedded DB.
https://rancher.com/docs/k3s/latest/en/installation/ha-embedded/

When multiple masters are specified in the master group, k3s-ansible will add the necessary flags during the initialization phase (i.e. --cluster-init and --server)

For the embedded HA mode to work the k3s version must be >= v1.19.1

Closes #32

PR #174, opened 2022-02-08
feat(loadbalance): added comments on how to use an external loadbalancer for HA

alteredtech:
Added commented out examples on how to use external loadbalancer with this play.

PR #210, opened 2023-02-01
Add support for HA cluster using embedded etcd

bubylou:
Configure K3S with more than one server with embedded DB so that with 3 servers 1 can go down while the cluster remains up. Mostly stolen from #97 although I ended up moving things around quite a bit. I tried keeping the HA setup simple but ended up making a number of other changes. Other changes include:

- Add reboot playbook for testing HA
- Update ansible modules with full names
- Copy kube config to localhost
- Change from master/node to server/agent naming scheme
- Switch inventory system to single inventory.yml file
- Moved playbooks to separate directory

Looking for feedback and testing. I've been running this against my K3S arm cluster with 3 server and 2 agent nodes.


bubylou:
Also added the ability to change K3S server location and changed the reset role to match which should fix #160. Adding HA mode also resolves #127. I could possibly add on a external DB option to close #32 as well.

bubylou:
Looking through other PRs mine has similar changes to #191, #192, #174, #131, and #50.

#166 has similar goals but I tried to keep my changes small so I didn't not include any load-balancing/VIP features and only mentioned them in the README.

bubylou:
Setting the k3s_server_location does not work properly unless set to one of the default locations which are /var/lib/rancher/k3s or ~/.rancher/k3s. To get around this I used a symbolic link which seems to work. It may just be my cluster but when I ran the playbook with a non-default k3s_server_location it ran fine but I had some kube-system pods crashing. A reboot of the whole cluster fixed it so I'm hoping it's just me. I'd be curious if anyone else can try this out to confirm if it works.

https://github.com/k3s-io/k3s/issues/5456
https://github.com/k3s-io/k3s/issues/1258

PR #218, opened 2023-04-11
Remove unsupported command module "warn" parameter

bcbrookman:
As of ansible-core 2.14.0, the builtin command module no longer has a "warn" parameter. Using it causes a fatal error that stops playbook completion.


bcbrookman:
The fatal error message is:

```
fatal: [k3s-master-01]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (ansible.legacy.command) module: warn. Supported parameters include: _raw_params, _uses_shell, argv, chdir, creates, executable, removes, stdin, stdin_add_newline, strip_empty_ends."}
```

dereknola (approved):

PR #224, opened 2023-09-12
FIX full k3s path in k3s-ansible/roles/k3s/master/tasks/main.yml

meilichios:
FIX full k3s path in k3s-ansible/roles/k3s/master/tasks/main.yml
Signed-off-by: Vittorio Distefano <info@vdistefano.studio>


dereknola:
Fixed with https://github.com/k3s-io/k3s-ansible/pull/143

Issue #221, opened 2023-06-28
Last instruction in README.md didn't make sense


daugherr:
I noted this in the cmemory issue but am logging a separate issue to make sure it's seen. 

The last instruction in README.md didn't make sense to me (scp debian@master_ip:~/.kube/config ~/.kube/config). 

kube.cfg was copied by Ansible to the master node. What didn't make sense to me was the difference between source and target. As written, wouldn't that copy in the same folder? I didn't do that step. Haven't deployed any containers yet but kubectl get nodes sees all my nodes just fine.

I feel like I'm missing something.

ninjaoxygen:
The `scp` command should be run locally on your workstation.

The intent is to copy the kube config from the master node to your local machine, so `kubectl` works from your machine, not just when you are `ssh`ed into the nodes.

daugherr:
Thanks! I do have a Linux laptop so I'll do that there. My primary workstation is still Windows 10. I haven't done much with the Linux subsystem. Will kubectl work from there or is more software needed? I have ssh set up using Powershell and using public/private keys only and a custom port vs. port 22 if that makes any difference. I've set up a config file under ../.ssh for that.

ninjaoxygen:
`scp` and `ssh` are both included in Windows 10. It also works via WSL under Win10/Win11 but that needs different configuration to Windows. I use PuTTY / pscp / plink / Pageant, so I'm not sure how to get the included Windows ones working.
The docs seem to say you can put your private keys at `C:\Users\<username>\.ssh\id_rsa` a.k.a `%USERPROFILE%\.ssh\id_rsa` and it will work.
`kubectl` I have installed from `scoop`, or get it from the official site at  https://kubernetes.io/docs/tasks/tools/install-kubectl-windows/

PR #228, opened 2023-11-08
Fix HA, simplify provisioning, add Vagrant test cluster

dereknola:
Changes:
- Simplified install:
  - Use the K3s install script to download the correct version of K3s, no need to have seperate steps per Architecture. Also enables the use of k3s-killall.sh and k3s-uninstall.sh scripts for future playbooks
  - Reworked service files to be up-to-date. Expanded number to include cluster-init and secondary server options.
  - Removed temporary k3s-init node. Now we just launch the service correctly the first time
  - Require the use of a user defined `token`. This greatly simplifies startup flow, as secondary nodes don't need to extract the randomly generated token from the inital node. Also allows use of ansible vault to provide an encrypted token value. 
- While initial code for HA was landed in #210, it did not acutally work for mutli server setups. HA is now working correctly.
- Add Vagrantfile for local testing. This provisions a 5 node cluster of ubuntu VMs, allowing local testing of the ansible playbooks. 
- Warn for several linting rules instead of failing

Resolves Issues:
- https://github.com/k3s-io/k3s-ansible/issues/221

Signed-off-by: Derek Nola <derek.nola@suse.com>


bubylou:
@dereknola what issues were there with the previous HA setup? I'm running a bare metal K3S cluster using that role and wanted to know if I needed to fix anything. I like your method of setting the token from the start better, much simplier and easier to maintain.

dereknola:
Hey @bubylou, The issue I was seeing was with the hostvars properly propagating to the other nodes, so agent-0 would be unable to join.

Testing on commit https://github.com/k3s-io/k3s-ansible/commit/9ecdc933ca279f1a4243a203e0432be6c7d9ef22, right before I switched to defined `token`, calling `vagrant up server-0 agent-0` (So a sqlite server with a single agent) I get the following error:
```
TASK [k3s/agent : Copy K3s service file] ***************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: ansible.errors.AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'token'. 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'token'
fatal: [agent-0]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'token'. 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'token'"}
```

With the hostvar section in question being:
```
--token {{ hostvars[groups['server'][0]]['token'] }}
```
Inside the k3s-agent.service.j2 file

Its possible that this is a Vagrant inventory specific issue, but after about an hour of debugging I was never able to get the other nodes to see the hostvar token from server-0.

PR #232, opened 2023-11-09
Fixed playbook structure

heindrichpaul:
* The roles couldn't be found when it sat at the root


dereknola:
This is not typically ansible project structures. You don't just nest everything in a single folder.

dereknola:
See https://github.com/techno-tim/k3s-ansible for another K3s ansible project structure.

PR #262, opened 2023-11-30
Fixed #260: k3s autocompletion is added to .bashrc only when necessary;

### LINKED ISSUE #260: "Add K3s autocomplete to user bashrc" does not test first if it is needed
Subsequent running of the [playbook/site.yml](https://github.com/k3s-io/k3s-ansible/blob/master/playbook/site.yml) will continue add k3s autocompletion to the shell profile.  It should test first if it is already there.

https://github.com/k3s-io/k3s-ansible/blob/1e8bfb0d3967defe66929616f50a2f40d2470a87/roles/k3s_server/tasks/main.yml#L78

And, the last line of the 3-line addition does not have an EOL ('\n').  But, this is an issue for k3s.

jon-stumpf:
k3s auto-completion is added _~/{{ ansible_user }}/.bashrc_ only when necessary. I replaced the use of **ansible.builtin.command** with **ansible.builtin.lineinfile**. This fixes https://github.com/k3s-io/k3s-ansible/issues/260.

### DIFF
--- a/roles/k3s_server/tasks/main.yml
+++ b/roles/k3s_server/tasks/main.yml
@@ -75,13 +75,11 @@
         owner: "{{ ansible_user }}"
         mode: "u=rw,g=,o="
 
-    - name: Add K3s autocomplete to user bashrc
-      become: true
-      become_user: "{{ ansible_user }}"
-      ansible.builtin.command:
-        cmd: "k3s completion bash -i"
-      register: out
-      changed_when: out.rc != 0
+    - name: Add K3s autocomplete to ~{{ ansible_user }}/.bashrc
+      ansible.builtin.lineinfile:
+        path: "~{{ ansible_user }}/.bashrc"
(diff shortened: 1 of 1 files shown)


roles/k3s_server/tasks/main.yml
```diff
@@ -75,13 +75,11 @@
...
         mode: "u=rw,g=,o="
 
-    - name: Add K3s autocomplete to user bashrc
-      become: true
-      become_user: "{{ ansible_user }}"
-      ansible.builtin.command:
-        cmd: "k3s completion bash -i"
-      register: out
-      changed_when: out.rc != 0
+    - name: Add K3s autocomplete to ~{{ ansible_user }}/.bashrc
```
dereknola, line 78:
```suggestion
    - name: Add K3s autocomplete to user bashrc
```

Original name avoid lint errors

dereknola (changes requested):

dereknola (approved):

PR #275, opened 2023-12-11
Fix k3s_server role (9998f503b4211211c46e4fef63d242d1ceb831ec)

UninspiredNickname:
9998f503b4211211c46e4fef63d242d1ceb831ec uses `&&` instead of `and` in `when` clause thus breaking play


dereknola:
Closing in favor of https://github.com/k3s-io/k3s-ansible/pull/273

PR #277, opened 2023-12-12
Allow empty `ansible_user`

pallxk:
#### Changes ####

Allow `ansible_user` to be specified as an empty string.

`ansible_user` specifies the username to use when connecting to hosts, but it should not be required to have a value in my opinion as the same info may also

* be provided in `~/.ssh/config`;
* be omitted if it happens to be the same username on both local and remote;
* be unnecessary if using local connection.

#### Linked Issues ####

None.

### DIFF
--- a/inventory-sample.yml
+++ b/inventory-sample.yml
@@ -12,7 +12,7 @@ k3s_cluster:
   # Required Vars
   vars:
     ansible_port: 22
-    ansible_user: debian
+    ansible_user: ""
     k3s_version: v1.26.9+k3s1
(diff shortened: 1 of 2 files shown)


dereknola:
This seems like an unnecessary complication. To support a local configuration, you would already need to modify the `playbook/site.yml` to have `connection: local` to support a `localhost` host value. At that point, why not just modify your inventory to also include the `ansible_user: MY_LOCAL_USER`. This change doesn't really help with a better "default case" as further modification to code is still required.

PR #280, opened 2023-12-16
add config playbook

roivanov:
Re-apply system configuration setting playbook added. User may have a case when a node was added manually to the cluster. And so user didn't pay close attention applying some/most of the configuration settings. This change should allow bringing such nodes in-sync with those configured by the script initially while skipping K3s installation steps.

#### Linked Issues ####

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,14 @@ A playbook is provided to upgrade K3s on all nodes in the cluster. To use it, up
 ansible-playbook playbook/upgrade.yml -i inventory.yml
 ```
 
+## Re-applying system configuration
+
+A playbook is provided to re-apply system configuration settings of the nodes in your cluster:
+
+```bash
+ansible-playbook playbook/config.yml -i inventory.yml
+```
+
 ## Airgap Install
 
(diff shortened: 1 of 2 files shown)


dereknola:
Can you explain your reasoning for this PR? Running the default `site.yaml` will always run the system configuration steps. Are you attempting to only apply the system steps, make changes before installation of K3s, then apply the full install?

roivanov:
> Can you explain your reasoning for this PR? Running the default `site.yaml` will always run the system configuration steps. Are you attempting to only apply the system steps, make changes before installation of K3s, then apply the full install?

User may have a case when a node was added manually to the cluster. And so user didn't pay close attention applying some/most of the configuration settings. This change should allow bringing such nodes in-sync with those configured by the script initially while skipping K3s installation steps.

dereknola:
This can already be accomplished with the default playbook. You just need to add an additional group to your k3s_cluster in your inventory. 
Here is an example:
```
k3s_cluster:
  children:
    server:
      hosts:
        192.168.1.91:
    agent:
      hosts:
        192.168.1.93:
    manual:
      hosts:
        192.168.1.92:
 ```

The name of the subgroup doesn't matter, as long as its not `server` or `agent`. We run the [Cluster-Prep](https://github.com/k3s-io/k3s-ansible/blob/master/playbook/site.yml#L2) on every target in the k3s_cluster

**Comment by dereknola** [comment]
Making a dedicated inventory of `manual.yml` would prevent you from rerunning k3s installation on you existing inventory.
```
k3s_cluster:
  children:
    manual:
      hosts:
        192.168.1.92:
```

Issue #257, opened 2023-11-28
Errors encountered when airgap_dir is undefined


jon-stumpf:
I attempted to create a simple cluster with three nodes.  Here is my **inventory.yml** file.
```
---
k3s_cluster:
  children:
    server:
      hosts:
        10.140.240.20:
    agent:
      hosts:
        10.140.240.21:
        10.140.240.22:

  vars:
    ansible_port: 22
    ansible_user: pi
    k3s_version: v1.26.9+k3s1
    token: "mytoken"  # Use ansible vault if you want to keep it secret
    api_endpoint: "{{ hostvars[groups['server'][0]]['ansible_host'] | default(groups['server'][0]) }}"
    extra_server_args: ""
    extra_agent_args: ""
```
When running _ansible-playbook playbook/site.yml_, I get the following errors from **role/airgap**:
```
TASK [airgap : Distribute K3s amd64 images [Airgap]] **************************************************************************************************************************************************
Tuesday 28 November 2023  12:35:25 -0500 (0:00:00.191)       0:00:44.666 ****** 
fatal: [10.140.240.20]: FAILED! => {"msg": "No file was found when using first_found. Use errors='ignore' to allow this task to be skipped if no files are found"}
fatal: [10.140.240.21]: FAILED! => {"msg": "No file was found when using first_found. Use errors='ignore' to allow this task to be skipped if no files are found"}
fatal: [10.140.240.22]: FAILED! => {"msg": "No file was found when using first_found. Use errors='ignore' to allow this task to be skipped if no files are found"}
```

This error will occur at the following locations:
- https://github.com/k3s-io/k3s-ansible/blob/1e8bfb0d3967defe66929616f50a2f40d2470a87/roles/airgap/tasks/main.yml#L60
- https://github.com/k3s-io/k3s-ansible/blob/1e8bfb0d3967defe66929616f50a2f40d2470a87/roles/airgap/tasks/main.yml#L75
- https://github.com/k3s-io/k3s-ansible/blob/1e8bfb0d3967defe66929616f50a2f40d2470a87/roles/airgap/tasks/main.yml#L90

I can add **ignore_errors: true** to the **ansible.builtin.copy** tasks to allow the playbook to proceed, but this does not seem the correct thing to do (given the error messages still produced).  And, since **airgap_dir** is not defined in my case, I don't understand why it is in the block to begin with.  See
[...]

PR #282, opened 2023-12-24
Update required ansible version

dann1:
#### Changes ####

Updated ansible minimum required version

#### Linked Issues ####

I ran into [this one](https://github.com/k3s-io/k3s-ansible/issues/257)

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ on processor architectures:
 
 ## System requirements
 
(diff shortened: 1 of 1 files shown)


dereknola:
Its seems that 2.15.6 also works, so I would prefer to reference the older version in the README

dann1:
done

Issue #292, opened 2024-01-28
Add the ability to set firewall rules for custom CIDR ranges


laszlojau:
In the current implementation, the CIDR ranges set in the firewall rules are hard-coded: `10.42.0.0/16` and `10.43.0.0/16`.
Users are able to override these CIDR ranges via the `cluster-cidr` and `service-cidr` flags/variables but this does not update the ranges used for the firewall rules.

PR #293, opened 2024-01-28
Set firewall rules for custom CIDR ranges

laszlojau:
#### Changes ####

role: prereq

#### Linked Issues ####

#292

### DIFF
--- a/roles/prereq/tasks/main.yml
+++ b/roles/prereq/tasks/main.yml
@@ -57,9 +57,7 @@
       community.general.ufw:
         rule: allow
         src: '{{ item }}'
-      loop:
-        - 10.42.0.0/16  # Pods
-        - 10.43.0.0/16  # Services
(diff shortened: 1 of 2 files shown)


laszlojau:
I tried to keep multiple options open for passing the CIDR range to the prereq role, happy to update/simplify the logic if needed.

dereknola:
I'm happy to support different options. But you should simplify the logic. Just have a `cluster_cidr` and `service_cidr` default vars that could be overriden by the inventory var. 

Additionally, you need to sign all your commits to comply with the DCO.

dereknola:
Yeah i think that would work:
- If hard defined as a variable, use that.
- Otherwise attempt to figure it out from the k3s config.yaml values 
- Otherwise use the defaults

laszlojau:
Updated the logic and signed the commit, let me know if you need anything else.

dereknola (approved):

Issue #264, opened 2023-11-30
[Epic] k3s_server and k3s_agent tasks are not idempotent


jon-stumpf:
The below tasks are repeated even when they are not needed.  Fixes are forthcoming.
### [k3s_server](https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_server/tasks/main.yml)
- [x] Download K3s binary, https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_server/tasks/main.yml#L13
- [x] Copy K3s service file [Single], https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_server/tasks/main.yml#L26
- [x] Copy config file to user home directory, https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_server/tasks/main.yml#L70
- [x] Add K3s autocomplete to user bashrc (#262), https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_server/tasks/main.yml#L78
- [x] Change server to API endpoint instead of localhost, https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_server/tasks/main.yml#L86
### [k3s_agent](https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_agent/tasks/main.yml)
- [x] Download K3s binary, https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_agent/tasks/main.yml#L13
- [x] Copy K3s service file (#268), https://github.com/k3s-io/k3s-ansible/blob/19c206d0cb448a33c869bbb8e084653e111be582/roles/k3s_agent/tasks/main.yml#L23

dereknola:
I assume most of this was me blindly putting in `changed_when: true` for these tasks :flushed:

dereknola:
GH being dumb and autoclosing this issue.

alvsanand:
[...]

PR #297, opened 2024-02-09
Check K3s installed version before download tasks

LawiK974:
#### Changes ####

- [Agent : Download artefact only if needed](roles/k3s_agent/tasks/main.yml#L13)
- [Server : Download artefact only if needed](roles/k3s_server/tasks/main.yml#L13)
- [Upgrade : Upgrade node only if needed](roles/k3s_upgrade/tasks/main.yml#L14)

#### Linked Issues ####

 #264  : k3s_server and k3s_agent tasks are not idempotent

### DIFF
--- a/roles/k3s_server/tasks/main.yml
+++ b/roles/k3s_server/tasks/main.yml
@@ -1,23 +1,38 @@
 ---
-# If airgapped, all K3s artifacts are already on the node.
-- name: Download K3s install script
-  when: airgap_dir is undefined
-  ansible.builtin.get_url:
-    url: https://get.k3s.io/
-    timeout: 120
-    dest: /usr/local/bin/k3s-install.sh
-    owner: root
-    group: root
-    mode: 0755
+- name: Get k3s installed version
+  ansible.builtin.command: k3s --version
+  register: k3s_version_output
+  changed_when: false
+  ignore_errors: true
 
-- name: Download K3s binary
-  when: airgap_dir is undefined
-  ansible.builtin.command:
-    cmd: /usr/local/bin/k3s-install.sh
-  environment:
-    INSTALL_K3S_SKIP_START: "true"
-    INSTALL_K3S_VERSION: "{{ k3s_version }}"
-  changed_when: true
+- name: Set k3s installed version
+  when: k3s_version_output.rc == 0
+  ansible.builtin.set_fact:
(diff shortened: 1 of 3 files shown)


roles/k3s_server/tasks/main.yml
```diff
@@ -1,23 +1,36 @@
...
-  when: airgap_dir is undefined
-  ansible.builtin.command:
-    cmd: /usr/local/bin/k3s-install.sh
-  environment:
-    INSTALL_K3S_SKIP_START: "true"
-    INSTALL_K3S_VERSION: "{{ k3s_version }}"
-  changed_when: true
+# If installed_k3s_version < k3s_version no need to reinstall binary
+- name: Download artefact only if needed 
+  when: k3s_version_output.rc != 0 or installed_k3s_version is version(k3s_version, '<') and airgap_dir is undefined
```
dereknola, line 16: Your comment is wrong, but your code is right. If the installed version of K3s on the nodes is older than the requested version in the config, we should be downloading and installing the newer version.

roles/k3s_upgrade/tasks/main.yml
```diff
@@ -1,34 +1,47 @@
...
-  changed_when: true
+- name: Set k3s installed version
+  set_fact:
+    installed_k3s_version: "{{ k3s_version_output.stdout_lines[0].split(' ')[2] }}"
 
-- name: Restore K3s service
-  ansible.builtin.shell:
-    cmd: "mv /tmp/k3s*.service {{ systemd_dir }}/"
-  changed_when: true
+# If installed_k3s_version < k3s_version no need to update node binary
```
dereknola, line 13: Same thing, comment is incorrect compared to "when" block.

dereknola (changes requested):

roles/k3s_upgrade/tasks/main.yml
LawiK974: Sorry for the delay and thank you for the review, I change the comment for all 3 files in a commit amend.

dereknola (approved):

dereknola:
Please fix lint errors

LawiK974:
You're right, sry, I hadn't paid attention to linting.
I fixed both ansible-lint and yamllint errors 👍️.

PR #300, opened 2024-02-18
Enable skipping bootcmd verification in Raspberry PI

jlpedrosa:
#### Changes ####
Add `skip_rpi_cmdline` variable to enable skipping the mutation of `/boot/firmware/cmdline.txt`vin Ubuntu Raspberry.
#### Linked Issues ####

### DIFF
--- a/inventory-sample.yml
+++ b/inventory-sample.yml
@@ -29,7 +29,6 @@ k3s_cluster:
     # extra_manifests: [ '/path/to/manifest1.yaml', '/path/to/manifest2.yaml' ]
     # airgap_dir: /tmp/k3s-airgap-images
     # user_kubectl: true, by default kubectl is symlinked and configured for use by ansible_user. Set to false to only kubectl via root user.
-    # skip_rpi_cmdline: true # normal instalations of raspberry don't require this, when set it spiks the checks of
     # server_config_yaml:  |
       # This is now an inner yaml file. Maintain the indentation.
       # YAML here will be placed as the content of /etc/rancher/k3s/config.yaml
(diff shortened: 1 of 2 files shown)


dereknola:
You wording would suggest that this is no longer required for any version of Ubuntu installed on Raspberry Pi. Is this true?

dereknola:
It seems like we either need it, or don't. If its an issue of newer versions don't require it, we can built that in. Adding an additional config is not a solution I want to pursue.

jlpedrosa:
Hey @dereknola ! It got lost of the previous PR that closed due to commit signing. 

The reason for this is NOT a version change, is that I'm booting the PIs through network boot and there's no `/boot/firmware/cmdline.txt` in some instalations.

I've never worked with Ansible, not sure what are the options to achive the behaviour that works in both scenarios.

roles/raspberrypi/tasks/prereq/Ubuntu.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - name: Enable cgroup via boot commandline if not already enabled
+  when: skip_rpi_cmdline is undefined
```
dereknola, line 3:
```suggestion
  when: lookup('fileglob', '/boot/firmware/cmdline.txt', errors='warn') | length == 0
```

dereknola: This will solve your problem of no file without failing. No need for an additional config option.

inventory-sample.yml
```diff
@@ -29,6 +29,7 @@ k3s_cluster:
     # extra_manifests: [ '/path/to/manifest1.yaml', '/path/to/manifest2.yaml' ]
     # airgap_dir: /tmp/k3s-airgap-images
     # user_kubectl: true, by default kubectl is symlinked and configured for use by ansible_user. Set to false to only kubectl via root user.
+    # skip_rpi_cmdline: true # normal instalations of raspberry don't require this, when set it spiks the checks of
```
dereknola, line 32:
```suggestion
```

dereknola (changes requested):

Issue #313, opened 2024-03-29
Add ability to define custom cluster context name in kubeconfig instead of 'k3s-ansible'


vrsarin:
**Issue**:

When automating fewer clusters, it is easier to manage cluster context manually.  We need to replace the hardcoded k3s-ansible cluster context name for multiple clusters with a user-defined context name.

**Proposed Solution**:
Create a variable in k3s-server main.yml defaults that can be overridden in inventory.yml when deploying.

PR #315, opened 2024-03-29
Added custom context name

vrsarin:
#### Changes ####
Added a variable called cluster_context to replace hardcoded cluster context name

#### Linked Issues ####
- #313

### DIFF
--- a/roles/k3s_server/tasks/main.yml
+++ b/roles/k3s_server/tasks/main.yml
@@ -133,14 +133,14 @@
           when: kubeconfig != "~/.kube/config"
           ansible.builtin.shell: |
             TFILE=$(mktemp)
-            KUBECONFIG={{ kubeconfig }} kubectl config set-context {{ cluster_context }} --user={{ cluster_context }} --cluster={{ cluster_context }}            
+            KUBECONFIG={{ kubeconfig }} kubectl config set-context {{ cluster_context }} --user={{ cluster_context }} --cluster={{ cluster_context }}
             KUBECONFIG={{ kubeconfig }} kubectl config view --flatten > ${TFILE}
             mv ${TFILE} {{ kubeconfig }}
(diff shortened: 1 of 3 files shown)


dereknola:
@vrsarin Please address the lint errors. Otherwise LGTM.

vrsarin:
> @vrsarin Please address the lint errors. Otherwise LGTM.

Fixed the linitng issues. Please approve.

roles/k3s_server/tasks/main.yml
```diff
@@ -120,20 +120,21 @@
...
           delegate_to: 127.0.0.1
           become: false
 
         - name: Merge with any existing kubeconfig on control node
           when: kubeconfig != "~/.kube/config"
           ansible.builtin.shell: |
             TFILE=$(mktemp)
-            KUBECONFIG={{ kubeconfig }} kubectl config set-context k3s-ansible --user=k3s-ansible --cluster=k3s-ansible
+            KUBECONFIG={{ kubeconfig }} kubectl config set-context {{cluster_context}} --user={{cluster_context}} --cluster={{cluster_context}}
+            KUBECONFIG={{ kubeconfig }} kubectl config use-context {{cluster_context}}
```
dereknola, line 137:
```suggestion
```

I don't want to change the context in case the user is already working with another kubeconfig context. We should try and touch the host configuration as little as possible. Thoughts?

dereknola (changes requested):
I pushed a lint fix. It was actually just complaining about a empty space at the end of line 137.

roles/k3s_server/tasks/main.yml
vrsarin: Agreed! This can be removed from here and added to the workflow as a task.

vrsarin:
@dereknola removed the line. Please approve.

PR #322, opened 2024-04-04
feat: Add support using Docker runtime & allow setting of node name per node

shanedell:
#### Changes ####

Add support using Docker runtime.
Allow setting of node name per node.

#### Linked Issues ####

- No linked issue, these were just somethings I noticed come up when testing.
  - When passing `--docker` for `extra_server_args` and `extra_agent_args` it causes issues as docker was not installed.
  - When using multiple hosts there was no easy to set the node name per node.

### DIFF
--- a/inventory-sample.yml
+++ b/inventory-sample.yml
@@ -38,3 +38,4 @@ k3s_cluster:
       # Containerd can be configured to connect to private registries and use them to pull images as needed by the kubelet.
       # YAML here will be placed as the content of /etc/rancher/k3s/registries.yaml
       # See https://docs.k3s.io/installation/private-registry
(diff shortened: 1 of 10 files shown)


dereknola:
The intent for this ansible playbook is as stated:

> https://github.com/k3s-io/k3s-ansible?tab=readme-ov-file#need-more-features
> This project is intended to provide a "vanilla" K3s install

That means using the default CRI containerd. This repo is not intended to exercise every possible option that K3s provides. If you want to make major modifications to it, feel free to fork the repo. 

Also if you want to run k3s under docker I would highly recommend looking at:

https://k3d.io/

vrsarin (changes requested):
Please see my suggestions.

PR #323, opened 2024-04-04
Fix some lints and firewalld policies

### LINKED ISSUE #321: ansible lint, playbooks folder, and firewalld
1. playbook folder should be renamed to playbooks - `ansible-playbook k3s.orchestration.site` will not works without it .
2. ansible-lint - please fix warings or add ignore comment `# noqa: ...`
3. Zone trusted accept traffic to all ports. What is the purpose of these tasks?
```
    - name: If firewalld enabled, open api port
      ansible.posix.firewalld:
        port: "{{ api_port }}/tcp"
        zone: trusted
        state: enabled
        permanent: true
        immediate: true

    - name: If firewalld enabled, open etcd ports
      when: groups['server'] | length > 1
      ansible.posix.firewalld:
        port: "2379-2381/tcp"
        zone: trusted
        state: enabled
        permanent: true
        immediate: true
```
It will be ok for 'zone: internal'. Then public IPs of cluster nodes can be added to 'internal' zones.
4. token and other parameters should be 'k3s_' prefixed

psztoch:
#### Linked Issues ####

Partially close #321


dereknola:
You will need to sign all commits submitted. See https://github.com/k3s-io/k3s-ansible/pull/323/checks?check_run_id=23445098776

roles/prereq/tasks/main.yml
```diff
@@ -82,10 +82,25 @@
...
         state: enabled
         permanent: true
         immediate: true
+      with_items:
+        - 8472/udp
+        - 10250/tcp
+        - 51820/udp
+        - 51821/udp
+        - 5001/tcp
+        - 6443/tcp
```
dereknola, line 103:
```suggestion
    - name: If firewalld enabled, open inbound ports
      ansible.posix.firewalld:
        port: "{{ item }}"
        zone: internal
        state: enabled
        permanent: true
        immediate: true
      with_items:
        - 8472/udp
        - 10250/tcp
        - 51820/udp
        - 51821/udp
        - 5001/tcp
```

You don't need 6443, its the `api_port` opened above.

dereknola (changes requested):

roles/prereq/tasks/main.yml
```diff
@@ -82,10 +82,25 @@
...
+
+    - name: If firewalld enabled, open inbound ports
+      ansible.posix.firewalld:
+        port: "{{ item }}"
+        zone: internal
         state: enabled
         permanent: true
         immediate: true
+      with_items:
+        - 8472/udp
```
vrsarin, line 98: Should these be left for the user to provide?

vrsarin (changes requested):
Please see my comments.

roles/prereq/tasks/main.yml
psztoch: Yes. It is my mistake. Sorry.

roles/prereq/tasks/main.yml
```diff
@@ -73,7 +73,7 @@
     - name: If firewalld enabled, open api port
       ansible.posix.firewalld:
         port: "{{ api_port }}/tcp"
-        zone: trusted
+        zone: internal
```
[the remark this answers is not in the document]
psztoch: @vrsarin Do you mean fw zone for inter-node traffic?

roles/prereq/tasks/main.yml
psztoch: In the internal zone, only the nodes would be added. It is not a big risk to open ports to different network backends. I believe this will generate less hassle for users.

roles/prereq/tasks/main.yml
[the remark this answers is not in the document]
vrsarin: If I am right, Linux allows the creation of a custom zone, which could be different from an OS-defined zone.

Do you know if hardcoding will benefit you here?

Today, you want to change it to internal. Tomorrow, I may want to revert back to trusted or my custom zone.

roles/prereq/defaults/main.yml
```diff
@@ -1,2 +1,13 @@
...
+# Zone for inter-node traffic
+k3s_firewalld_node_zone: internal
+
+# List of IP addresses or cidr masks of your nodes
+k3s_firewalld_node_cidrs: []
+
+# List of public services
+k3s_firewalld_public_ports:
+  - 80/tcp
+  - 443/tcp
```
dereknola, line 11: Why are we exposing public ports, these aren't used by K3s?

Honestly this whole rework of firewalld has me questioning even trying to keep network configuration in the playbook. Everyone's network is going to be different. Everyone clearly has different ideas about what they want. We should just point people to the network requirements in the docs and say "preconfigure your network with your own ansible playbook, then use this to install K3s"

roles/prereq/tasks/main.yml
vrsarin: In my opinion, we should give users the flexibility to define their ports or exclude the port configuration from the scope of this role.

vrsarin: Yeah, agreed, The user can manage adding or removing ports in a separate task.

roles/prereq/defaults/main.yml
dereknola, line 11: K3s does not require any public ports to function, so we should not have any set as default. Only user provided ports will be opened. 

```suggestion
# List of public services
k3s_firewalld_public_ports: []
```

dereknola (changes requested):

dereknola:
@psztoch Did you still want to handle this PR, or are you going to just carry patches in your fork?

dereknola:
Closing in favor of https://github.com/k3s-io/k3s-ansible/pull/329 and https://github.com/k3s-io/k3s-ansible/pull/326

Issue #327, opened 2024-05-04
Hard Coded Group name for k3s servers


fkonradmain:
As already mentioned in #321 , some variable names are not ideal and should be prefixed.

> 4. token and other parameters should be 'k3s_' prefixed

The host group names for "server" and "agent" are not ideal either. The name is ambiguous. Additionally it is a host variable that is also active on other plays in the same playbook. It could possibly cause collisions with them.

Hence, in my playbook I renamed the server group to "k3s_server" and the agent group "k3s_agent". While this works for the agent playbook, it does not work for the server playbook. Unfortunately, the group names are hard coded into the latter.

This issue comes up, every time the following call is made:
```
groups['server']
```

My idea would be to introduce a variable, e.g. `inventory_server_group_name` and have that name default to `server`. In that case we would just have to change every occurence of `groups['server']` to:

```
groups['{{ inventory_server_group_name }}']
```

I am willing to implement this solution myself and provide it as pull request, yet I would like to hear the admin's opinion first.

dereknola:
Hi @fkonradmain, I understand your issue, but could use your help better understanding "how" you and potentially other users utilize this playbook in concert with others. My day to day use of large scale ansible playbooks is limited. I typically am using only 1 or 2 separate playbooks on an inventory.

Because of my use case and potentially other beginners, when I took over as maintainer, I didn't see a reason change all the vars and the group names and add `k3s_` it seemed redundant to me. You are using the k3s-ansible playbook, of course all the stuff in the inventory is for K3s. 

In the K3s context, `server` and `agent` are the official names we give node roles. The naming inside the K3s space is not ambiguous, but very clear and used throughout the k3s-io org and docs.

I'm trying to balance the clarity and ease-of-use for first time ansible users (K3s is so easy to deploy that we often get people with no experience in Kubernetes starting with K3s) and people who may be utilizing this playbook in large scale enterprise deployments. 

What does a typically inventory look like for you? Why is `server` and `agent` to ambiguous? What other playbooks do you utilize that come in conflict with this one?  Would the groups being changed to `k3s_server` and `k3s_agent` be enough to resolve your issue?

simagick:
Currently the server role will try to join servers with the first node in the server group to create the control plane.  This means that we can not  manage multiple clusters within the same inventory with the hard-coded group names. 

Making configurable groups gives us more flexibility.   I have a PR that addresses this issue in #331

PR #331, opened 2024-06-01
Make agent and server groups configurable

simagick:
#### Changes ####
Inventory group names `server` and `agent` are  configured through `server_group` and `agent_group` variables.  

The use of hard-coded group names causes issues when managing multiple clusters within the same inventory.    When the control plane is initialized each server node connects to the first member of the server group, and therefore all members of each `server` group must belong to the same cluster.

This change allows more flexibility in how operators manage their inventory.  A configurable `agent` group is added for completeness.

#### Linked Issues ####
https://github.com/k3s-io/k3s-ansible/issues/327


roles/k3s_upgrade/tasks/main.yml
```diff
@@ -54,14 +54,14 @@
...
-      when: "'server' in group_names"
+      when: "server_group in group_names"
       ansible.builtin.systemd:
         state: restarted
         daemon_reload: true
         name: k3s
 
     - name: Restart K3s service [agent]
-      when: "'agent' in group_names"
+      when: "server_agent in group_names"
```
dereknola, line 64: Typo?
```suggestion
      when: "agent_group in group_names"
```

dereknola (changes requested):

dereknola:
Lint is failing for unrelated reason, I'll handle in another PR

PR #334, opened 2024-06-08
Add group variables to the prereq role

laszlojau:
#331 added support for overriding the `server` and `agent` group names in the `k3s_server` and `k3s_upgrade` roles, but the same changes were not added to the `prereq` role.

#### Changes ####
This PR makes the `server` and `agent` groups configurable in the `prereq` role as well.

#### Linked Issues ####
#327


dereknola:
Unrelated lint failure

dereknola (approved):

PR #339, opened 2024-06-20
POC: Supporting k3s-ansible with external database

peterklijn:
#### Changes ####

A (working) proof of concept of supporting external databases in this Ansible setup.
This change introduces an option variable `use_external_database`, which if present uses the `k3s-single.service.j2` template, regardless of how many server hosts are defined.
Using this together with the `extra_server_args` variable to pass along a `--datastore-endpoint` will create a K3s setup using an external datastore.

```yaml
---
k3s_cluster:
  children:
    server:
      hosts:
        192.168.64.2:
        192.168.64.3:
    agent:
      hosts:
        192.168.64.4:

  # Required Vars
  vars:
    # ...
    token: mytoken
    use_external_database: true
    extra_server_args: "--datastore-endpoint=postgres://username:password@hostname:port/database-name"
    # ...
```

#### Linked Issues ####
https://github.com/k3s-io/k3s-ansible/issues/340


roles/k3s_server/tasks/main.yml
```diff
@@ -58,7 +58,7 @@
   when: inventory_hostname == groups[server_group][0]
   block:
     - name: Copy K3s service file [Single]
-      when: groups[server_group] | length == 1
+      when: groups[server_group] | length == 1 or use_external_database | default(false) | bool
```
dereknola, line 61: Adding default value of `false` to https://github.com/k3s-io/k3s-ansible/blob/master/roles/k3s_server/defaults/main.yml would simplify your when statements.

dereknola (changes requested):

dereknola:
If you rebase off master, you won't see the lint error around ansible 2.14 anymore.

roles/k3s_server/tasks/main.yml
peterklijn: Thanks!
I've updated the PR

dereknola (approved):

peterklijn:
Hi @dereknola,
Thanks for the review, I've fixed the lint error and added some documentation to the readme.
Could you have another look and let me know if there's anything I need to do before merging?

roles/k3s_server/tasks/main.yml
```diff
@@ -58,7 +58,7 @@
   when: inventory_hostname == groups[server_group][0]
   block:
     - name: Copy K3s service file [Single]
-      when: groups[server_group] | length == 1
+      when: groups[server_group] | length == 1 or use_external_database | bool
```
dereknola, line 61:
```suggestion
      when: groups[server_group] | length == 1 or use_external_database
```

roles/k3s_server/tasks/main.yml
```diff
@@ -67,7 +67,9 @@
         mode: "0644"
 
     - name: Copy K3s service file [HA]
-      when: groups[server_group] | length > 1
+      when:
+        - groups[server_group] | length > 1
+        - use_external_database | bool == false
```
dereknola, line 72:
```suggestion
        - not use_external_database
```

roles/k3s_server/tasks/main.yml
```diff
@@ -148,14 +150,27 @@
     - inventory_hostname != groups[server_group][0]
   block:
     - name: Copy K3s service file [HA]
-      when: groups[server_group] | length > 1
+      when:
+        - groups[server_group] | length > 1
+        - use_external_database | bool == false
```
dereknola, line 155:
```suggestion
        - not use_external_database
```

roles/k3s_server/tasks/main.yml
```diff
@@ -148,14 +150,27 @@
...
         src: "k3s-ha.service.j2"
         dest: "{{ systemd_dir }}/k3s.service"
         owner: root
         group: root
         mode: "0644"
 
+    - name: Copy K3s service file [External DB]
+      when:
+        - groups[server_group] | length > 1
+        - use_external_database | bool
```
dereknola, line 166:
```suggestion
        - use_external_database
```

dereknola (changes requested):
Sorry for the delay, was offsite all last week. This should fix all the lint. As `use_external_database` is already defined with a default bool value, all attempts to override it must also be a bool, eliminating the need to filter `| bool` in the when clauses.

peterklijn:
> Sorry for the delay, was offsite all last week. This should fix all the lint. As `use_external_database` is already defined with a default bool value, all attempts to override it must also be a bool, eliminating the need to filter `| bool` in the when clauses.

Thanks for the review! I've processed your comments 👍

PR #344, opened 2024-07-06
Add a handler to restart the K3s Server when the service file changes

peterklijn:
#### Changes ####

Introduced a handler to restart the K3s service when a change to the `{{ systemd_dir }}/k3s.service` is detected.

I ran into the issue several times where I changed something in the template or `extra_server_args` and did not see my change reflected, and had to restart the service manually to see the change applied.
This PR solves that problem

### DIFF
--- a/roles/k3s_server/tasks/main.yml
+++ b/roles/k3s_server/tasks/main.yml
@@ -65,6 +65,7 @@
         owner: root
         group: root
         mode: "0644"
+      notify: Restart K3s Server
 
     - name: Copy K3s service file [HA]
       when:
@@ -76,6 +77,7 @@
         owner: root
         group: root
         mode: "0644"
+      notify: Restart K3s Server
 
     - name: Add service environment variables
       when: extra_service_envs is defined
@@ -159,6 +161,7 @@
         owner: root
         group: root
         mode: "0644"
+      notify: Restart K3s Server
 
     - name: Copy K3s service file [External DB]
       when:
@@ -170,6 +173,7 @@
         owner: root
         group: root
         mode: "0644"
+      notify: Restart K3s Server
 
     - name: Enable and check K3s service
       ansible.builtin.systemd:
--- a/roles/k3s_server/tasks/main.yml
+++ b/roles/k3s_server/tasks/main.yml
@@ -65,7 +65,7 @@
         owner: root
         group: root
         mode: "0644"
-      notify: Restart K3s Server
+      register: service_file_single
 
     - name: Copy K3s service file [HA]
       when:
@@ -77,7 +77,7 @@
         owner: root
         group: root
         mode: "0644"
-      notify: Restart K3s Server
+      register: service_file_ha
 
     - name: Add service environment variables
       when: extra_service_envs is defined
@@ -86,7 +86,18 @@
         line: "{{ item }}"
       with_items: "{{ extra_service_envs }}"
 
+    - name: Restart K3s service
+      when:
+        - ansible_facts.services['k3s.service'] is defined
+        - ansible_facts.services['k3s.service'].state == 'running'
+        - service_file_single.changed or service_file_ha.changed
+      ansible.builtin.systemd:
+        name: k3s
+        daemon_reload: true
+        state: restarted
+
     - name: Enable and check K3s service
+      when: ansible_facts.services['k3s.service'] is not defined or ansible_facts.services['k3s.service'].state != 'running'
       ansible.builtin.systemd:
         name: k3s
         daemon_reload: true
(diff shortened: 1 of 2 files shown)


roles/k3s_server/tasks/main.yml
```diff
@@ -65,6 +65,7 @@
         owner: root
         group: root
         mode: "0644"
+      notify: Restart K3s Server
```
dereknola, line 68: Did you test this on a fresh install? Will this cause problems with "restarting" a service that has never been enabled or started before? On first install we don't enable K3s service for another 20lines or so.

peterklijn: Handlers are run after all tasks in a play complete, so it shouldn't cause a problem.
In the case of a clean install, it will restart in the end even though it's not necessary, but it shouldn't hurt it either.

dereknola: I don't like that we are always restarting the service, even if there are no changes. My suggestion is to check if the K3s service us already running. If it is, restart the K3s service.


So instead of  
```
    - name: Enable and check K3s service
      ansible.builtin.systemd:
        name: k3s
        daemon_reload: true
        state: started
        enabled: true
```

We have:
```
    - name: Check if K3s service is already running
      ansible.builtin.systemd:
        name: k3s
        state: started
      register: k3s_status
      ignore_errors: true
      
    - name: Enable and start K3s service
      when: k3s_status.status != 'running'
      ansible.builtin.systemd:
        name: k3s
        daemon_reload: true
        state: started
        enabled: true

    - name: Restart existing K3s Service
      when: k3s_status.status == 'running'
      ansible.builtin.systemd:
        name: k3s
        daemon_reload: true
        state: restarted
```

dereknola: Just tested this and its actually `ansible_facts.services['k3s.service']`

peterklijn: > I don't like that we are always restarting the service, even if there are no changes. My suggestion is to check if the K3s service us already running. If it is, restart the K3s service. No handler needed.

I think there's some miscommunication, so just in case, let me explain:

A handler only runs when the task that notifies it is marked as changed. So when you run the playbook multiple times, it will not call the handler ([docs](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_handlers.html#notifying-handlers))
In this specific case:

```yaml
    - name: Copy K3s service file [Single]
      when: groups[server_group] | length == 1 or use_external_database
      ansible.builtin.template:
        src: "k3s-single.service.j2"
        dest: "{{ systemd_dir }}/k3s.service"
         owner: root
         group: root
         mode: "0644"
       notify: Restart K3s Server
```

This will only 'notify' the handler when the `/etc/systemd/system/k3s.service` file changes, for example:
- when the `k3s-single.service.j2` template changes
- when the `extra_server_args` is updated
- etc

The only catch is that the very first time you run this playbook to create a new server, ansible marks the file as changed as it creates it. So only the very first time does this run multiple times.
I assume these playbooks run many times to configure and update the cluster over time, and it's more important that it doesn't restart unnecessarily, so having a redundant restart only on the very first run is not a big deal.

Having said that, there's ways to prevent that, but I think that the complexity it introduces is not worth it.
The solution you provide is one way, but it does not have my preference as it is needed at multiple places, as the 'Enable and check service' is in two places ([1](https://github.com/k3s-io/k3s-ansible/blob/master/roles/k3s_server/tasks/main.yml#L87)) ([2](https://github.com/k3s-io/k3s-ansible/blob/master/roles/k3s_server/tasks/main.yml#L174))

So, does the above change your mind?
And if not, do you have a strong preference on using regular tasks instead of handlers?

dereknola: I do have a strong preference for using regular tasks. I am fine with the code being repeated in two places. We treat secondary servers differently enough to warrant it. 

For some more context, I use this tool to quickly spin up and then kill K3s clusters for various QA and CI related work. So I am concerned with 1 additional restart because my use case is heavily dependent on startup time.

peterklijn: Gotcha, thanks for clarifying.
I'll update it in the upcoming days

peterklijn: @dereknola I've updated the PR, a couple small changes from your suggestion:
- The restart task is before the start, as otherwise it will always be started and restart as well
- Registered some variables so it only restarts when the service file is changed

I've tested it with and without changes, for both servers with no `k3s.service` running, and without. It works as expected 🙂

dereknola (approved):

PR #352, opened 2024-08-06
More flexible cgroup settings

anon-software:
Currently the the relevant tasks match the kernel command line to three cgroup related parameters in the hard-coded order. I already had two of them and when I ran the playbook I ended up with two of them being repeated. A cleaner way to apply it is to match one parameter at the time and add it if missing.

I tested only Debian version, but the other are a simple copy and paste except for Archlinux.


dereknola (approved):

anon-software:
I used a wrong module in non-Debian tasks. I shall push another commit with the correction.

vrsarin (approved):
approved

dereknola (approved):

PR #356, opened 2024-08-22
Security exposure related to the token

anon-software:
The installation playbook saves the token into the systemd unit configuration file /etc/systemd/system/k3s.service. The problem is that according to K3s' documentation "the server token should be guarded carefully" (https://docs.k3s.io/cli/token), yet the configuration file is readable by anybody. A better solution is to save the token into its corresponding environment file /etc/systemd/system/k3s.service.env which is readable by the super user only. This is what the standard K3s' installation script (https://get.k3s.io) does.

#### Linked Issues ####

### DIFF
--- a/roles/k3s_agent/tasks/main.yml
+++ b/roles/k3s_agent/tasks/main.yml
@@ -35,14 +35,13 @@
         INSTALL_K3S_EXEC: "agent"
       changed_when: true
 
-    - name: Add the token and first server URL for joining the cluster to the environment
+    - name: Add the token for joining the cluster to the environment
       no_log: true # avoid logging the server token
       ansible.builtin.lineinfile:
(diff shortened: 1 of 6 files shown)


dereknola (changes requested):
I don't think there is any reason to set the server URL as an env variable. I take your point around token security, but there isn't an issue with keeping `--server https://{{ api_endpoint }}:{{ api_port }}`

anon-software:
That is correct, the URL can remain where it is. It just looked nicer to me to move that as well. But I shall restore it where it was if you prefer the old location.

dereknola:
Please restore it :)

anon-software:
I have pushed another commit to move the URL back.

dereknola (approved):

PR #364, opened 2024-09-26
Prevent multiple tokens in k3s.service.env

anon-software:
If site.yml playbook is executed multiple times with different tokens, they will all accumulate in k3s.service.env. They won't do any harm because the last one wins, however it is a matter of good housekeeping to delete the old before inserting a new one.

#### Linked Issues ####


anon-software:
This is not ideal. When executed again with the same token, the playbook now deletes and inserts the same line resulting in the task reporting a "change" when in fact everything effectively remained the same. I working on a fix for this this.

dereknola (approved):

Issue #307, opened 2024-03-15
Make passing a token optional


jgooge-vso:
The RKE2 ansible automation will generate a token if one isn't explicitly defined. It'd be nice if this automation was configured the same way.

dereknola:
Please see https://github.com/k3s-io/k3s-ansible/issues/294, as this was already asked for and answered. I ran into a problem of bootstrapping later server nodes with the token generated on the first. More broadly, this repo is maintain by me as a low priority. Hopefully it works and doesn't have bugs, but I don't have the bandwidth to add new features. If you need other features, I am happy to review and merge PRs or I would suggest looking at the [Other Projects Shoutouts in the README](https://github.com/k3s-io/k3s-ansible?tab=readme-ov-file#need-more-features).

dereknola:
I continue to maintain that autogenerating a token is low-priority. Either you "don't care" about the token and so why not just make one up `token=myfavorite123` because the cluster is going to be short lived or security is not a priority. Or you really do care, in which case, you can use Vault or `--extra-vars token=$(rand or gpw or pwgen command)` to pass a secure token that does not exist in plaintext on your inventory.yaml.

PR #375, opened 2024-11-08
Generate token

anon-software:
If a token is not explicitly provided, let the first server generate a random one. Such a token is saved on the first server and the playbook can retrieve it from there and store it a a fact. All other servers and agents can use that token later to join the cluster. It will be saved into their environment file as usual.

I tested this by creating a cluster of one server and then adding two more servers and one agent. Please let me know if I should try some other tests as well.

#### Linked Issues ####
https://github.com/k3s-io/k3s-ansible/issues/307

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -28,6 +28,7 @@ def provision(vm, role, node_num)
     ansible.extra_vars = {
       k3s_version: "v1.28.14+k3s1",
       api_endpoint: "#{NETWORK_PREFIX}.100",
+      # Required for vagrant ansible provisioner
       token: "myvagrant",
       # Required to use the private network configured above
       extra_server_args: "--node-external-ip #{node_ip} --flannel-iface eth1", 
--- a/inventory-sample.yml
+++ b/inventory-sample.yml
@@ -19,6 +19,7 @@ k3s_cluster:
     # - openssl rand -base64 64
     # - pwgen -s 64 1
     # You can use ansible-vault to encrypt this value / keep it secret.
+    # Or you can omit it if not using Vagrant and let the first server automatically generate one.
     token: "changeme!"
     api_endpoint: "{{ hostvars[groups['server'][0]]['ansible_host'] | default(groups['server'][0]) }}"
     extra_server_args: ""
--- a/roles/k3s_agent/defaults/main.yml
+++ b/roles/k3s_agent/defaults/main.yml
@@ -1,4 +1,5 @@
 ---
+server_group: server  # noqa var-naming[no-role-prefix]
 k3s_server_location: "/var/lib/rancher/k3s"  # noqa var-naming[no-role-prefix]
 systemd_dir: "/etc/systemd/system"  # noqa var-naming[no-role-prefix]
 api_port: 6443  # noqa var-naming[no-role-prefix]
--- a/roles/k3s_agent/tasks/main.yml
+++ b/roles/k3s_agent/tasks/main.yml
@@ -35,6 +35,10 @@
         INSTALL_K3S_EXEC: "agent"
       changed_when: true
 
+- name: Get the token from the first server
+  set_fact:
+    token: "{{ hostvars[groups[server_group][0]].token }}"
+
 - name: Delete any existing token from the environment if different from the new one
   ansible.builtin.lineinfile:
     state: absent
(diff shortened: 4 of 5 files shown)


dereknola:
This has been tried before. You need to test the case of `3 servers` all at the start. There are issues with the fact correctly being propagated to the other two servers, as the bring up on all the nodes happens asynchronously, so you cannot guarantee (and when I tried to implemented this it failed) that the fact will even exist for the other 2 servers to see and join with.

anon-software:
Maybe I do not understand how Ansible works then. Doesn't the task "Init first server node" from roles/k3s_server/tasks/main.yml runs first and terminates before "Start other server if any and verify status" runs? The former task will save the token and it will always be available for the others being setup in the latter, or at least that is how thought it would work.

Setting up three servers all at once was actually the first test I ran, although it is possible that it was a fluke that it worked.

dereknola:
If you got it working, that great! I'm gonna pull down your PR and check it out sometime later today or Monday.

dereknola:
When testing with the vagrant file, I see the following error
```
TASK [k3s_agent : Get the token from the first server] *************************
fatal: [agent-0]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'token'\n\nThe error appears to be in '/home/derek/rancher/ansible-k3s/roles/k3s_agent/tasks/main.yml': line 38, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Get the token from the first server\n  ^ here\n"}
```

dereknola:
So interesting results. For the 3 pi cluster, the first time I tested with 3 servers, it installed fine. Then ran the `reset` playbook and tried to run the `site.yaml` again. This time it also failed with
```
TASK [k3s_server : Get the token from the first server] *******************************************************************************************************************************
fatal: [192.168.1.91]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'token'\n\nThe error appears to be in '/home/derek/rancher/ansible-k3s/roles/k3s_server/tasks/main.yml': line 208, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  block:\n    - name: Get the token from the first server\n      ^ here\n"}
fatal: [192.168.1.92]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'token'\n\nThe error appears to be in '/home/derek/rancher/ansible-k3s/roles/k3s_server/tasks/main.yml': line 208, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n  block:\n    - name: Get the token from the first server\n      ^ here\n"}

PLAY RECAP ****************************************************************************************************************************************************************************
192.168.1.90               : ok=21   changed=3    unreachable=0    failed=1    skipped=45   rescued=0    ignored=1   
192.168.1.91               : ok=21   changed=3    unreachable=0    failed=1    skipped=61   rescued=0    ignored=1   
192.168.1.92               : ok=21   changed=3    unreachable=0    failed=1    skipped=60   rescued=0    ignored=1   

```

This is the exact same issue I ran into the first time I attempted to implement auto generating tokens.

dereknola:
Run a server + agent inventory on the raspberry pi cluster, the playbook works because those are seperate roles, so they run sequentially (ie the server role gets executed, then the agent role). But for the vagarant provisioner, it just runs everything in parallel, so this system will never work. 

I'm less concerned if the Vagrantfile works, that can just be notes in the `Vagrantfile` as "requires token". But the above errors around regular ssh nodes is a blocker on this PR.

dereknola:
[cluster.log](https://github.com/user-attachments/files/17707647/cluster.log)

dereknola:
Okay nvmd I just read my own error logs. Let me fix it.

dereknola:
I seem to have found a seperate issue around `Copy K3s service file` needing `extra_server_args` to be defined. I had stripped down my inventory.yaml to be super simple. I will open a seperate PR to address this issue.

anon-software:
OK, I shall push another commit to address the new batch of Lint errors.

dereknola (changes requested):
Can you add a comment above https://github.com/k3s-io/k3s-ansible/blob/master/Vagrantfile#L31 That a token variable is required for the vagrant ansible provisioner. 

I'm happy to accept this PR if it work on "real" ansible playbooks and vagrant is just weird.

dereknola (approved):

Issue #368, opened 2024-10-07
Some file are not created (k3s.service.env, k3s-agent.service.env)


jkrenzer:
Hi!

Due to the missing parameter `create: true` in the `ansible.builtin.lineinfile` command, some files are not created when they are empty and thus an installation from a clean system fails.

So far I identified following steps:

In roles/k3s_server/tasks/main.yaml:
- Add service environment variables
- Add token as an environment variable

In roles/k3s_agent/tasks/main.yaml:
- Add the token for joining the cluster to the environment

Cheers,

Jörn

dereknola:
K3s-ansible is not responsible for creating the `k3s.service.env` file. That is created by the k3s install script. https://github.com/k3s-io/k3s/blob/master/install.sh#L961. What types of errors are you seeing?

jkrenzer:
First the errors were censored, when setting `no_log: false` in the aforementioned tasks it was a file not found error. 

System is a freshly installed Debian 11/bullseye server, ansible collection is from the master branch here.

sky003:
I probably have a similar issue. Perhaps it has something to do with this change https://github.com/k3s-io/k3s-ansible/commit/bc03d1c0d053bb3e3362c526aa38bc8762d1f816

This file is not created `{{ systemd_dir }}/k3s-agent.service.env`. And I don't see where it's supposed to be created by k3s. 

So probably we should test the file before running the task:
```
- name: Add the token for joining the cluster to the environment
  no_log: false # avoid logging the server token
  ansible.builtin.lineinfile:
    path: "{{ systemd_dir }}/k3s-agent.service.env"
    line: "{{ item }}"
  with_items:
    - "K3S_TOKEN={{ token }}"
  when: "'{{ systemd_dir }}/k3s-agent.service.env' is file"
```

dereknola:
I'm still unable to reproduce this issue, I can clearly find the k3s-agent.service.env on a fresh installed ubuntu node:
```
root@agent-0:/home/vagrant# cat /etc/systemd/system/k3s-agent.service.env 
K3S_TOKEN=myvagrant
```

dereknola:
Both agent and server service and  service.env files are created when the k3s install script runs on the nodes.

dereknola:
What does your inventory.yaml look like, are you changing any of the k3s configuration?

dereknola:
Ah I think I found the problem. On debian 11, the install script is not getting downloaded correctly
```
[...]

PR #385, opened 2025-01-06
Make sure we create token environment file if it does not exist already

albal:
As it currently stands, when the `lineinfile` tasks run in either agent or server roles it fails if the file does not exist.  I have added the create option set to true:
```yaml
create: yes
```

You can check the [Ansible documentation](https://docs.ansible.com/ansible/2.9/modules/lineinfile_module.html#parameter-create)


dereknola:
Those files are automatically created by the K3s install script and should always be present when that step is run.
https://github.com/k3s-io/k3s/blob/master/install.sh#L968

This is effectively https://github.com/k3s-io/k3s-ansible/issues/368 again. Please check your setup and make sure you are not attempting to install both servers and agents on the same node. What does your ansible inventory look like?

PR #399, opened 2025-03-20
fix: make the roles work even when hosts are aliased in the inventory

### LINKED ISSUE #397: Obtaining randomly generated token from the first server fails when using aliases in the Ansible intentory
When using alias in Ansible inventory `inventory_hostname` is different from what's in `group['server']`. Usnig `ansible_host` should fix that.

MikkCZ:
#### Changes ####
I made it so that the roles are using `ansible_host` instead of the `inventory_hostname`. That way the lookup into the list of hosts works even if the inventory hosts have an alias https://docs.ansible.com/ansible/latest/inventory_guide/intro_inventory.html#inventory-aliases.

#### Linked Issues ####
#397

### DIFF
--- a/roles/airgap/tasks/main.yml
+++ b/roles/airgap/tasks/main.yml
@@ -97,7 +97,7 @@
           skip: true
 
     - name: Run K3s Install [server]
-      when: inventory_hostname in groups['server']
+      when: inventory_hostname in groups['server'] or ansible_host in groups['server']
       ansible.builtin.command:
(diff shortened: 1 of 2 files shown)


pawlakus:
This PR has a potential impact and it's not backwards compatible. why not add your condition as alternative instead? I would rather try this: 

when: inventory_hostname == ... or ansible_host == ...

MikkCZ:
That makes sense. Done.

dereknola (approved):

PR #406, opened 2025-04-20
Fix bug: When the k3s_server_location variable is defined in the conf…

ghostelement:
…iguration, the deployment process fails on certain systems if the directory specified by k3s_server_location does not exist.

#### Changes ####
roles/prereq/tasks/main.yml
#### Linked Issues ####


dereknola:
It is assumed that if you want a custom k3s data dir, you are handling its creation and setup prior to running this playbook, since its likely that your rolling a special OS that requires non standard install locations.

PR #407, opened 2025-04-20
Support ufw use sources IP set rules

ghostelement:
#### Changes ####
roles/prereq/tasks/main.yml
add:  Support ufw use sources IP set rules
#### Linked Issues ####

### DIFF
--- a/roles/prereq/tasks/main.yml
+++ b/roles/prereq/tasks/main.yml
@@ -29,6 +29,19 @@
 - name: Populate service facts
   ansible.builtin.service_facts:
 
+# Set ufw_internal_sources IPs
+- name: Set UFW sources IPs
+  ansible.builtin.set_fact:
+    ufw_internal_sources: >-
+      {{
+        (
+          groups[server_group] | default([])
+          + groups[agent_group] | default([])
+        )
+        | map('extract', hostvars, ['ansible_default_ipv4', 'address'])
+        | flatten | unique | list
+      }}
+
 - name: Allow UFW Exceptions
   when:
     - ansible_facts.services['ufw'] is defined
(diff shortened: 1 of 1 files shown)


roles/prereq/tasks/main.yml
```diff
@@ -64,6 +81,25 @@
         rule: allow
         src: '{{ item }}'
       loop: "{{ (cluster_cidr + ',' + service_cidr) | split(',') }}"
+    
+    - name: If ufw enabled, allow default ports
+      when:
+        - ufw_status['stdout_lines'][0] == "Status: active"
+      block:
```
dereknola, line 88: Move the above 3 checks inside the block. No need to keep checking `"'Status: active' in ufw_status['stdout']"`

dereknola (changes requested):

dereknola:
Right now, the bare-minimum ports are opened for k3s to work on ufw. That's why we are effectively forcing these exceptions onto ubuntu nodes. However, the additional ports and IP are not required for K3s to work. These "additional" exceptions should be placed under a new configuration bool `ufw_exceptions` (or something similar).

ghostelement:
Thank you for your response. I’ve realized that my previous description was problematic. In fact, in the old code, the ufw firewall policy allowed all visitors to access the k3s ports without restricting the source addresses. So, in my branch, I added a list of source addresses. Now I’ve further improved the code to allow these rules to be added even when ufw is enabled, rather than only when it is in the active state. I believe this is more in line with security requirements.

roles/prereq/tasks/main.yml
```diff
@@ -41,29 +54,42 @@
       register: ufw_status
```
dereknola, line 54: We don't need this task, now that we don't check `ufw_status` on any of the subsequent tasks.

dereknola:
@ghostelement all commits need to be signed to comply with the DCO set by CNCF. See https://github.com/k3s-io/k3s-ansible/pull/407/checks?check_run_id=40907593281 for more info.

dereknola (changes requested):

PR #411, opened 2025-05-18
ansible: Install wget and curl for k3s server download

behroozam:
These tools are required to fetch the K3s installation script during setup.
Here is the error message Set up k3s ansible on a host without `wget` or `curl` .
image

#### Changes ####
Adding wget and curl to allow downloading the k3s binary, 
#### Linked Issues ####

### DIFF
--- a/roles/prereq/tasks/main.yml
+++ b/roles/prereq/tasks/main.yml
@@ -258,4 +258,4 @@
     name:
       - wget
       - curl
-    state: latest
+    state: present
--- a/roles/prereq/tasks/main.yml
+++ b/roles/prereq/tasks/main.yml
@@ -252,3 +252,10 @@
         content: "{{ registries_config_yaml }}"
         dest: "/etc/rancher/k3s/registries.yaml"
         mode: "0644"
+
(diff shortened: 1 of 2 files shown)


behroozam (commented):
Nice suggestion. 
Thank you

roles/prereq/tasks/main.yml
```diff
@@ -252,3 +252,10 @@
         content: "{{ registries_config_yaml }}"
         dest: "/etc/rancher/k3s/registries.yaml"
         mode: "0644"
+
+- name: Install wget and curl required for downloading the k3s binary
```
dereknola, line 256:
```suggestion
- name: Install wget and curl required for downloading the k3s binary
  when: airgap_dir is undefined
```

dereknola: I realized that we need to ensure this check is not run on airgapped environments as they do not use wget/curl.

dereknola (changes requested):

behroozam:
> Closing as stale.

Hi, I updated my branch. 
Can you please open the PR again 
Thanks

PR #414, opened 2025-05-25
Fix: add ufw exceptions if rules already exist

movd:
#### Changes ####

Before, the whole stdout output of `ufw status` had to be `'Status: active.'` With ufw rules already present, this leads to the condition not getting fulfilled. 

Now only the first line has to be `Status: active` for the k3s-ansible rules to get applied.

#### Linked Issues ####

N/A


dereknola:
So this is a fix partially reverting https://github.com/k3s-io/k3s-ansible/pull/403? 

I don't run ufw, so I never tested any of these ufw changes.

dereknola:
Please fix your PR so the link check passes.

Issue #424, opened 2025-07-27
Error adding new k3s agent


daisaacson:
Get en error at the `Add the token for joining the cluster to the environment` step

with `no_log` disabled, here is the error:

```
TASK [k3s_agent : Add the token for joining the cluster to the environment] *************************************************************************************************************************
failed: [k3s-3.local] (item=K3S_TOKEN=mysecret) => {"ansible_loop_var": "item", "changed": false, "item": "K3S_TOKEN=mysecret", "msg": "Destination /etc/systemd/system/k3s-agent.service.env does not exist !", "rc": 257}
```

ansible-playbook version
```
$ ansible-playbook --version
ansible-playbook [core 2.18.7]
  config file = None
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/linuxbrew/.linuxbrew/Cellar/ansible/11.8.0/libexec/lib/python3.13/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/linuxbrew/.linuxbrew/bin/ansible-playbook
  python version = 3.13.5 (main, Jun 11 2025, 15:36:57) [GCC 11.4.0] (/home/linuxbrew/.linuxbrew/Cellar/ansible/11.8.0/libexec/bin/python)
  jinja version = 3.1.6
  libyaml = True
```

daisaacson:
here are results with pull request [425](https://github.com/k3s-io/k3s-ansible/pull/425)

```
TASK [k3s_agent : Add the token for joining the cluster to the environment] *************************************************************************************************************************
changed: [k3s-3.local] => (item=None)
changed: [k3s-3.local]
```

The ansible script continued to process the role without further issues.

dereknola:
I'm unable to replicate your issue with the nodes I have available:
```
k3s_cluster:
  children:
    server:
      hosts:
        192.168.1.90:
        noder2.local:
        noder4.local:
    agent:
      hosts:
        noder3.local: # Pi 3B+
  vars:
    ansible_user: "pi"
    k3s_version: v1.33.3+k3s1
    api_endpoint: "{{ hostvars[groups['server'][0]]['ansible_host'] | default(groups['server'][0]) }}"
    token: "mysecret"
    extra_server_args: ""
    extra_agent_args: ""
```

[...]

PR #425, opened 2025-07-27
create file if it doesn't exist

daisaacson:
#### Changes ####
add `create` parameter to `ansible.builtin.lineinfile` task.
This allows the file to be created if it doesn't exist.

#### Linked Issues ####
[424](https://github.com/k3s-io/k3s-ansible/issues/424)


daisaacson:
When I encountered this error 2 weeks ago, the playbook couldn't recover/continue from the error state I fell in. Re-running the playbook multiple times would result in the same error. The only thing that got the playbook to continue was the updates in this pull request.
Maybe that's an argument for this pull request, but since the error state is not reproducible by others or myself, I will close this pull request.

Issue #427, opened 2025-08-05
Airgap: Copy multiple image archives


Claiyc:
As a SWE I want the airgap image distribution mechanism to copy ALL image archives in {{ airgap_dir }} to /var/lib/rancher/k3s/agent/images/.

This allows the injection of custom images during k3s bootstrap.

PR #428, opened 2025-08-05
Distribute multiple image archives

Claiyc:
#### Changes ####
Add "Gather image archives" ansible finder step
Modify "Distribute K3s images" ansible step to distribute all archives found in previous step

#### Linked Issues ####
#427

### DIFF
--- a/roles/airgap/tasks/main.yml
+++ b/roles/airgap/tasks/main.yml
@@ -81,22 +81,17 @@
         mode: "0755"
         state: directory
 
-    - name: Gather image archives
-      find:
-        paths: "{{ airgap_dir }}"
-        patterns: "*.tar*"
-      register: found_images
-      delegate_to: localhost
-      become: false
-
-    - name: Distribute image archives
+    - name: Distribute Airgap images {{ k3s_arch }}
       ansible.builtin.copy:
-        src: "{{ item.path }}"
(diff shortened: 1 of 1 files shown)


roles/airgap/tasks/main.yml
```diff
@@ -81,21 +81,22 @@
...
-    - name: Distribute K3s images {{ k3s_arch }}
+    - name: Gather image archives
+      find:
+        paths: "{{ airgap_dir }}"
+        patterns: "*.tar*"
+      register: found_images
+      delegate_to: localhost
+      become: false
+
+    - name: Distribute image archives
```
dereknola, line 92: You don't need to break this out into two steps, this can be accomplished with a slight modification to the original copy, while ensuring non images don't get placed and cause issues.

```suggestion
    - name: Distribute Airgap images {{ k3s_arch }}
      ansible.builtin.copy:
        src: "{{ item }}"
        dest: /var/lib/rancher/k3s/agent/images/{{ item | basename }}
        owner: root
        group: root
        mode: "0755"
      with_fileglob:
        - "{{ airgap_dir }}/*.tar.gz"
        - "{{ airgap_dir }}/*.tar.zst"
        - "{{ airgap_dir }}/*.tar"
```

dereknola (changes requested):

dereknola (approved):

PR #453, opened 2025-10-21
Add: swapoff tasks in prereq role

cod3mas0n:
#### Changes ####

- Add disabling swap task
  - disable_swap task in prereq role
  - add handlers for daemon_reload after removing swap from `/etc/fstab`
  - include the disable_swap task in `main.yml`

### DIFF
--- a/roles/prereq/handlers/main.yml
+++ b/roles/prereq/handlers/main.yml
@@ -1,0 +1,4 @@
+---
+- name: Reload systemd daemon
+  ansible.builtin.systemd_service:
+    daemon_reload: true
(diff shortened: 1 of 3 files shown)


dereknola:
That's a good callout. However, more recent version of K8s no longer have as heavy of a suggestion to disable swap. Thats because starting with 1.28 beta and then GA in 1.32+, swap support is quite robust. See https://kubernetes.io/blog/2025/03/25/swap-linux-improvements/

I am actually in favor of removing the README section on swap being disabled. Thats out of date info.

cod3mas0n:
That's Awesome, investing significant effort into supporting swap.
Thank you.

Then I close this PR.

Issue #459, opened 2025-11-07
Missing inter-node ports for UFW


rpressiani:
Currently, the playbook handles inter-node communication ports (such as those for Flannel, Kubelet, and Spegel) correctly when Firewalld is enabled, but it doesn't open the equivalent set of ports when UFW is enabled.

PR #460, opened 2025-11-07
feat: add ufw allow inter-node ports

rpressiani:
#### Changes ####
Add a new task within the Allow UFW Exceptions block to open inter-node ports, mirroring the logic used in the Firewalld section.

#### Linked Issues ####
#459


rpressiani:
Closes #460

PR #474, opened 2025-12-05
feat(k3s_upgrade): regenerate service files from templates instead of restoring backups

lexfrei:
#### Changes ####

The `k3s_upgrade` role currently saves service files to `.bak` before upgrade and then restores them afterward. This means any changes to `extra_server_args` or `extra_agent_args` in the Ansible inventory are **not applied** during upgrades - the old service configuration is always restored.

This PR changes the upgrade role to regenerate service files from templates (like `k3s_server` and `k3s_agent` roles do) instead of restoring backups. This ensures that:

1. Configuration changes in Ansible variables are applied during upgrades
2. Ansible remains the single source of truth for K3s configuration
3. The upgrade role behavior matches the initial deployment roles

**Changes:**
- Replace backup restoration with template-based service file regeneration
- Add `force_reconfigure` flag to allow config-only changes without version upgrade
- Reuse templates from `k3s_server` role via symlink
- Add necessary defaults for template variables

**⚠️ Potential Breaking Change:**

This changes the upgrade behavior. Previously, manual modifications to service files on nodes would persist through upgrades. With this change, service files are regenerated from Ansible templates, which aligns with infrastructure-as-code principles but may surprise users who made manual changes.

However, this behavior was never documented, and it's reasonable to expect Ansible to be the source of truth. Users who need custom service configurations should define them via `extra_server_args`/`extra_agent_args` variables.

**Usage:**

```bash
# Normal upgrade (applies both version and config changes)
ansible-playbook playbooks/upgrade.yml

# Config-only update (no version change required)
ansible-playbook playbooks/upgrade.yml -e force_reconfigure=true
```

#### Linked Issues ####

None

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,75 @@
+---
+name: Integration Test
+
+on:
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        container_os: [debian12]
+
+    # K3s requires privileged containers to run inside Docker and access to cgrougs.
+    steps:
+      - name: Checkout codebase
+        uses: actions/checkout@v6
+
+      - name: Set up Python 3.13.
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.13.x'
+
+      - name: Install Ansible and dependencies
+        run: |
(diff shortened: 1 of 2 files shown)


mbathe19:
If merged the following PR isn't needed anymore https://github.com/k3s-io/k3s-ansible/pull/473

Regards

PR #482, opened 2025-12-19
Improve config file/service file rollout in k3s_server role

ppascente:
#### Changes ####
- I noticed the k3s_server config file update and k3s service restart behavior in the k3s_server role was sometimes causing a cluster outage if the config file was malformed or if a bad parameter was passed.
- This seemed to stem from 1. a fact getting set about the running state of k3s service no longer being valid further down the progression of the play (after something gets restart for instance) and 2. the task to check and enable the service getting skipped after checking that stale fact
- Some of those conditions are unnecessary as a result of how the idempotency built into the ansible systemd module works

Originally I had written this as a refactor to ensure the service was getting checked (and failing the play if a successful restart did not occur)
```
    - name: Restart K3s service
      when:
        - ansible_facts.services['k3s.service'] is defined
        - service_file_single.changed or service_file_ha.changed or _server_config_result.changed
      ansible.builtin.systemd:
        name: k3s
        daemon_reload: true
        state: restarted

    - name: Enable and check K3s service
      ansible.builtin.systemd:
        name: k3s
        state: started
        enabled: true
```
But I had noticed the slick way this was handled in the k3s_agent role and decided to replicate that. 

**Finally**
I had run into an issue more than once when a bad config.yaml was deployed, the first system's k3s service was restarted, and it moved on to fail deploying / restarting the service on the rest of the server nodes all at once, leaving the cluster in an non-operational state. This is likely fixed by the systemd service check working now, but I included a throttle to rollout config file changes and service restarts to the remaining servers one-by-one. Maybe this is overkill and/or we don't want a play to fail if one of many servers fails to restart gracefully. 

#### Removed ####
This PR includes a variable in the k3s_server and k3s_agent roles called *upgrade_preferred*
- This variable is used to fail the site.yml playbook if there's a version mismatch between installed k3s version and k3s_version variable, and the preferred experience is to use the upgrade.yml playbook. 
- The primary use-case that prompted this is when running the playbook via a pipeline, I want to promote k3s_version between environments, but do not want it accidentally triggering a reinstall on a cluster using the site.yml playbook


ppascente:
Heard - I suppose this is a workaround for site.yml not really being declarative in nature for k3s versioning. Really I think upgrade and site should be one in the same eventually. 

I can take those bits out and keep the systemd fixes.

PR #505, opened 2026-01-30
Fix / Feature: Make server and agent role idempotent and merge upgrade logic into them

### LINKED ISSUE #485: site.yml playbook is not (safely) declarative
The behavior when running site.yml against a set of systems where k3s is already running could be improved to include the functionality of upgrade.yml. This would allow a single playbook to manage the cluster's versioning over time in a safe manner, rather than having to use upgrade.yml in a procedural way. 

Running site.yml against an existing cluster with a new version declared is a disruptive action.

Making improvements to 2 roles over time vs 3 roles over time might make upkeep simpler.

I propose including the k3s_upgrade role's functionality into the k3s_server and k3s_agent roles and removing the upgrade playbook/role.

wombelix:
#### Changes ####

This PR mainly addresses the topic that using `site` on an existing cluster can be disruptive especially in combination with upgrading the version. The majority of changes was moving existing logic from the upgrade role to the server and agent role. With some minor adjustments and fixes. I also adjusted the GitHub Action integration test to validate that a re-run is idempotent and an upgrade through `site` instead of `upgrade` work. README is updated as well.

I personally prefer this way, re-run the same playbook for all cases (install, config changes, upgrades) instead of having one for the initial setup and another for upgrades. I understand that this was / is a controversial topic and no clear yes or no decision was made. Maybe this PR helps to make the decision easier ;) Worked in a single node setup, I would rely on the GHA integration test for HA to validate the changes.

Currently the upgrade playbook and role are not touched or removed, wasn't sure if they should stay for a while for backward compatibility or not.

#### Linked Issues ####

Fixes: https://github.com/k3s-io/k3s-ansible/issues/485

### DIFF
--- a/playbooks/site.yml
+++ b/playbooks/site.yml
@@ -11,6 +11,7 @@
 - name: Setup K3S server
   hosts: server
   become: true
+  serial: 1
   roles:
     - role: k3s_server
 
--- a/playbooks/site.yml
+++ b/playbooks/site.yml
@@ -11,7 +11,6 @@
 - name: Setup K3S server
   hosts: server
   become: true
-  serial: 1
   roles:
     - role: k3s_server
 
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -82,12 +82,22 @@ jobs:
         if: matrix.service_mgr == 'openrc'
         run: docker exec agent-node rc-service k3s-agent status | grep started
 
+      - name: Re-run site.yml to verify idempotency
+        env:
+          ANSIBLE_FORCE_COLOR: '1'
+        run: ansible-playbook playbooks/site.yml -i tests/basic.yml
+
+      - name: Verify K3s is still running after idempotent re-run
+        run: docker exec server-node k3s kubectl get nodes | grep Ready
+
       - name: Modify the k3s_version in inventory for upgrade
         run: |
           sed -i 's/k3s_version: v1.33.4+k3s1/k3s_version: v1.34.1+k3s1/' tests/basic.yml
 
-      - name: Run Upgrade Playbook
-        run: ansible-playbook playbooks/upgrade.yml -i tests/basic.yml
+      - name: Upgrade via site.yml
+        env:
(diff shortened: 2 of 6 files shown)


dereknola:
Thanks for the PR. 

> Currently the upgrade playbook and role are not touched or removed, wasn't sure if they should stay for a while for backward compatibility or not.

The removal of the upgrade playbook would constitute a major change to this playbook, likely resulting in a `v2.0.0` release. So immediate removal of the upgrade playbook is likely out of scope for the foreseeable future.

However, making the site.yaml handle all scenarios seems like the best hybrid. If people like the "smaller" upgrade only playbook, it still exists for use. Others such as yourself, wombelix, can have the "single playbook that does everything". I don't think the overhead of supporting both will be onerous.

playbooks/site.yml
```diff
@@ -11,6 +11,7 @@
 - name: Setup K3S server
   hosts: server
   become: true
+  serial: 1
```
dereknola: This is breaking CI and testing. Why do we need to run this serially? This should not be a requirement. The n+1 servers wait anyways for the first server to startup.

dereknola (changes requested):

lexfrei:
I hit the same gap this PR targets — a version-only bump applied through `site.yml` leaves agent nodes (and joined servers) on the old runtime — and wanted to help get the fix over the line.

Since this branch was parked pending the move off the jinja templates, and that refactor has since largely landed, I reworked the change on top of the now template-free roles rather than rebasing this one: https://github.com/k3s-io/k3s-ansible/pull/544. It keeps the `serial: 1` rolling behaviour for the etcd quorum, adds a per-server health gate, and the integration tests now upgrade through `site.yml` and assert the running (kubelet-reported) version so they fail if the restart regresses.

Your original work is credited as `Co-authored-by` on the commit. Happy to fold in any feedback — and of course if you'd rather carry it forward yourself, just say and I'll step back.

dereknola:
Closing in favor of #544

PR #528, opened 2026-04-07
HA joiner servers get undefined token error and generate their own random cluster secret when token is not set in inventory

mattyo161:
## Description
When deploying an HA cluster (multiple hosts in server group) without an explicit token in inventory, additional server nodes (joiners) fail with an Ansible error and/or start k3s with a newly generated random token — creating a separate one-node cluster instead of joining the existing one.

### Bug 1 — combine({'token': token}) references undefined variable
In roles/k3s_server/tasks/main.yml, the "Start other server" block reads the join token from the first server's random_token fact:
```yaml
- name: Get token from first server if needed
  when: token is not defined and hostvars[groups[server_group][0]].random_token is defined
  ansible.builtin.set_fact:
    k3s_server_config: "{{ k3s_server_config | combine({'token': token}) }}"
```

The `when: condition` correctly guards against token being undefined — but the set_fact value references `token` directly, which is still undefined at this point. This causes an Ansible fatal error:
```
'token' is undefined
```
The intent is clearly to use `hostvars[groups[server_group][0]].random_token` (the value read from the first server's token file), not `token` (the inventory variable that is explicitly not set).

#### Fix:
```
k3s_server_config: "{{ k3s_server_config | combine({'token': hostvars[groups[server_group][0]].random_token}) }}"
```

### Bug 2 — lineinfile regexp references token without a default
In the same "Start other server" block, the env file cleanup task uses `token | regex_escape` without a `default()`:
```
regexp: "^K3S_TOKEN=\\s*(?!{{ token | regex_escape }}\\s*$)"
```
When token is undefined this raises a template error. The first server's equivalent task already uses `token | default('') | regex_escape` — the joiner task should be consistent.

#### Fix:
```
regexp: "^K3S_TOKEN=\\s*(?!{{ token | default('') | regex_escape }}\\s*$)"
```
With `default('')`, when token is undefined the regexp becomes `(?!\s*$)` — which removes any non-empty `K3S_TOKEN=` line, preventing a stale env token from overriding the join token written to config.yaml.

## Steps to reproduce
1. Define 3+ hosts in the server group with no token in group_vars
2. Run k3s.orchestration.site targeting all server hosts
3. Observe fatal error on joiner hosts at Get token from first server if needed
    Alternatively, if the error is suppressed: inspect /etc/rancher/k3s/config.yaml on joiner nodes — token: will be absent and k3s will generate its own secret at startup, preventing cluster join.

## Version
- k3s.orchestration: 1.2.0
- Introduced by: [No server templates in k3s_server role #509](https://github.com/k3s-io/k3s-ansible/pull/509)

### DIFF
--- a/roles/k3s_server/tasks/main.yml
+++ b/roles/k3s_server/tasks/main.yml
@@ -271,7 +271,7 @@
       when: token is not defined and hostvars[groups[server_group][0]].random_token is defined
       # noqa var-naming[no-role-prefix]
       ansible.builtin.set_fact:
-        k3s_server_config: "{{ k3s_server_config | combine({'token': token}) }}"
+        k3s_server_config: "{{ k3s_server_config | combine({'token': hostvars[groups[server_group][0]].random_token}) }}"
 
     - name: Convert server config to YAML and write to file
       when: not ansible_check_mode
@@ -285,7 +285,7 @@
       ansible.builtin.lineinfile:
         state: absent
         path: "{{ k3s_server_env_file }}"
-        regexp: "^K3S_TOKEN=\\s*(?!{{ token | regex_escape }}\\s*$)"
+        regexp: "^K3S_TOKEN=\\s*(?!{{ token | default('') | regex_escape }}\\s*$)"
 
     - name: Reload systemd daemon
       when:
--- a/roles/k3s_upgrade/tasks/main.yml
+++ b/roles/k3s_upgrade/tasks/main.yml
@@ -12,6 +12,29 @@
     k3s_upgrade_current_version: "{{ k3s_upgrade_version_output.stdout_lines[0].split(' ')[2] }}"
   check_mode: false
 
+# Get the generated token if one is not defined in the configuration, we need to do this
+# outside of the next check in the case that the first server doesn't need the upgrade
+# this is same technique used in `k3s_server`
+# NOTE: this requires that the first serfver is included in any limited playbook run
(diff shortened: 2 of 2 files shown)


mattyo161:
I wanted to validate this so I updated the workflow in my fork and ran two tests:
1. mou-main - Integration Test - This is an updated integration test with code from the 1.2.0 release
2. mou-latest - Integration Test - This is an updated integration test with the updated code from this PR. 

As you can see in the [workflow run here](https://github.com/mattyo161/k3s-ansible/actions/runs/24260301912/job/70842150286#step:9:553) we get the following error when the `token` is not specified in the inventory and we rely on the random token generated by k3s when it is first installed on server node 1:
```
[ERROR]: Task failed: Finalization of task args for 'ansible.builtin.set_fact' failed: Error while resolving value for 'k3s_server_config': 'token' is undefined
```

When the same job is run against the `mou-latest` branch with the associated changes, it passes: https://github.com/mattyo161/k3s-ansible/actions/runs/24258508185/job/70836102063

dereknola:
@mattyo161 Appreciate the full detailed PR with testing, makes reviewing much easier. 

Please [sign your commits](https://github.com/k3s-io/k3s-ansible/pull/528/checks?check_run_id=70844178352) and push back to this PR, we need to comply with the CNCF DCO. 

After that I can approve and merge.

dereknola:
Closing as stale

Issue #374, opened 2024-11-07
k3s --version : Command not found


dheurtev:
When running the playbooks, I get an error k3s -- version : Command not found
Fix : In your roles source code, the version should be checked as /usr/local/bin/k3s --version instead of k3s --version
My system is Almalinux 9.

/k3s/orchestration/roles/k3s_agent/tasks/main.yml
3c3
<   ansible.builtin.command: k3s --version
---
>   ansible.builtin.command: /usr/local/bin/k3s --version
/k3s/orchestration/roles/k3s_server/tasks/main.yml
3c3
<   ansible.builtin.command: k3s --version
---
>   ansible.builtin.command: /usr/local/bin/k3s --version
246c246
<           k3s kubectl get nodes -l "node-role.kubernetes.io/control-plane=true" -o=jsonpath="{.items[*].metadata.name}"
---
>           /usr/local/bin/k3s kubectl get nodes -l "node-role.kubernetes.io/control-plane=true" -o=jsonpath="{.items[*].metadata.name}"
/k3s/orchestration/roles/k3s_upgrade/tasks/main.yml
5c5
<   ansible.builtin.command: k3s --version
---
>   ansible.builtin.command: /usr/local/bin/k3s --version

dereknola:
We cannot assume that the k3s install location is at `/usr/local/bin/k3s`. Depending on the operations system, the binary may be located else (this is often the case in read only/atomic OSes). The assumption for that line is that K3s will already have been installed on the system, and the PATH will include the location that k3s is in (as should be standard for most programs).


It is expected for that command to fail on instances where K3s has never been installed. Thats why we `ignore_errors: true` for that check.

dheurtev:
Understood, yet in [roles/k3s_upgrade/tasks/main.yml](https://github.com/k3s-io/k3s-ansible/blob/master/roles/k3s_upgrade/tasks/main.yml), ignore_errors:true is not set.
Shouldn't it be added there too ?

dereknola:
Your correct those are both problems with the current repo. I will open a PR to address them.

dereknola:
[...]

PR #543, opened 2026-07-03
fix: use failed_when instead of ignore_errors for version detection

archy-rock3t-cloud:
#### Changes ####
Use `failed_when: false` instead of `ignore_errors: true` for the "Get k3s installed version" probe in `k3s_server` and `k3s_agent`. The command is expected to exit non-zero when k3s isn't installed yet (see #374); `failed_when: false` tolerates it the same way (the following `rc == 0` check still works) but reports ok, so a successful first run no longer prints a red `FAILED!`/`ignoring` line.

#### Linked Issues ####
Refs #374

### DIFF
--- a/roles/k3s_agent/tasks/main.yml
+++ b/roles/k3s_agent/tasks/main.yml
@@ -3,7 +3,7 @@
   ansible.builtin.command: k3s --version
   register: k3s_agent_version_output
   changed_when: false
-  ignore_errors: true
+  failed_when: false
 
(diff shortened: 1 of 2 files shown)


archy-rock3t-cloud:
@dereknola the reason for this PR is on the screenshot. Red error in the middle of execution caused me to scroll up to make sure everything was ok more than once
red-error

dereknola (approved):

PR #544, opened 2026-07-04
fix: apply k3s version bumps declaratively through site.yml

lexfrei:
## What

Re-running `site.yml` with a raised `k3s_version` now upgrades the running cluster instead of only replacing the on-disk binary. Previously the install script replaced the binary but never restarted the service (`INSTALL_K3S_SKIP_START`), and the roles only restarted on a config change — so a version-only bump left agents and joined servers running the old runtime until they were restarted by hand.

## Why

This makes `site.yml` safely declarative for the cluster lifecycle, which is the ask in #485. It rebuilds the direction started in #505 on top of the now template-free roles: that earlier PR was parked pending the jinja-to-config.yaml refactor, which has since landed, so this is a fresh implementation rather than a rebase of it.

## How

- The server and agent roles track whether the k3s binary was actually replaced and restart when it was, in addition to the existing config-change trigger.
- The server play runs with `serial: 1` (matching `upgrade.yml`) and each restarted server — including the first — waits for its API to report healthy before the play moves on, so a rolling change never drops the etcd quorum.
- The joined-server readiness check waits for the servers handled so far rather than the full set, which keeps it correct under both serial and parallel runs.
- Integration tests upgrade through `site.yml` (single server + agent, and the three-server HA path) and assert the running version the kubelet reports, not the on-disk binary, so the tests fail if the restart ever regresses.

## Scope note

`upgrade.yml` and the `k3s_upgrade` role are intentionally left in place and unchanged. #485 also suggested removing them, but that is a larger, breaking change (a v2.0.0 concern) and is deliberately out of scope here — `site.yml` simply gains the declarative-upgrade capability alongside the existing playbook.

## Testing

The full integration suite — including the HA job's rolling `site.yml` version bump across three etcd servers — was run green on a fork before opening this PR.

Addresses #485. Builds on the approach in #505 (credited via `Co-authored-by` on the commit).

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -19,6 +19,7 @@ jobs:
     env:
       STARTING_K3S_VERSION: v1.34.3+k3s1
       UPGRADE_K3S_VERSION: v1.35.1+k3s1
+      SITE_UPGRADE_K3S_VERSION: v1.35.6+k3s1
     # K3s requires privileged containers to run inside Docker and access to cgrougs.
     steps:
       - name: Set container OS based on service manager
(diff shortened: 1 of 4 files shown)


.github/workflows/integration.yml
```diff
@@ -129,14 +129,31 @@ jobs:
         run: |
           sed -i "s/k3s_version: .*/k3s_version: ${UPGRADE_K3S_VERSION}/" tests/${{ matrix.inventory }}.yml
 
-      - name: Run Upgrade Playbook
```
dereknola, line 132: We still want to test the older upgrade playbook. People are still using it. Keep this here and test the second upgrade with the site.yml

roles/k3s_agent/tasks/main.yml
```diff
@@ -98,5 +109,5 @@
 - name: Enable and start K3s agent
   ansible.builtin.service:
     name: k3s-agent
-    state: "{{ 'restarted' if _agent_config_result.changed else 'started' }}"
+    state: "{{ 'restarted' if (k3s_agent_binary_changed or _agent_config_result.changed) else 'started' }}"
```
dereknola: Lets just simplify this and always make it `restarted` because that's guaranteed to bounce the service and handle both cases, then we don't need to do a bunch of logic around determining if the binary was replaced.

dereknola (changes requested):

.github/workflows/integration.yml
lexfrei: Restored the `upgrade.yml` test and added the `site.yml` bump as a second upgrade on top of it (start → `upgrade.yml` → `site.yml`, across three real k3s versions). The HA job already ran them in that order.

roles/k3s_agent/tasks/main.yml
lexfrei: Simplified to always `state: restarted` and dropped the binary-change detection. Applied the same simplification to the server role for consistency: the first server already always restarted, and joined servers now do too — still one at a time behind `serial: 1` + the health check, so the etcd quorum is preserved.

playbooks/site.yml
```diff
@@ -8,10 +8,14 @@
...
     - role: raspberrypi
 
+# Servers are handled one at a time so a rolling change (config edit or version
+# bump re-applied through this playbook) never restarts more than one server at
+# once and cannot break the etcd quorum. Agents have no such limitation.
 - name: Setup K3S server
   hosts: server
   gather_facts: false
   become: true
+  serial: 1
```
dereknola: We are not going to reduce the initial install of k3s to serial just to support upgrades. If you want this, use the `ansible-playbook site.yml --forks=1` system to force serial. Include that in the documentation.

dereknola (changes requested):

playbooks/site.yml
lexfrei: Done — dropped `serial: 1` from the server play, so a fresh install runs in parallel as before. The joined-server restart is now unconditional, which is the actual version-bump fix (the first server and the agents already always restart). The README documents running a multi-server upgrade with `--forks=1`, and the HA integration test now runs the `site.yml` upgrade with `--forks=1` so that path is covered.

dereknola (approved):

PR #550, opened 2026-08-01
Wrap api_endpoint with ansible.utils.ipwrap for IPv6 compatibility

sestegra:
#### Changes ####
Currently, `roles/k3s_server/tasks/main.yml` builds the kubeconfig server URL as:

```yaml
KUBECONFIG={{ kubeconfig }} kubectl config set-cluster default --server=https://{{ api_endpoint }}:{{ api_port }}
```

This assumes `api_endpoint` is either a hostname or an IPv4 address. When `api_endpoint` is a bare IPv6 address (e.g. `2001:db8::1`), the resulting URL is malformed:

```
https://2001:db8::1:6443
```

which is ambiguous/invalid.

So, we have to work around this by pre-bracketing the value themselves in inventory (`api_endpoint: "[2001:db8::1]"`), which is inconsistent with how `wait_for.host` and `tls-san` expect the same address elsewhere in the role (unbracketed).

**Fix**
Apply `ansible.utils.ipwrap` at the point of URL construction, so the inventory value stays a plain address/hostname everywhere, and only the URL-building template adds brackets when needed:

```yaml
KUBECONFIG={{ kubeconfig }} kubectl config set-cluster default --server=https://{{ api_endpoint | ansible.utils.ipwrap }}:{{ api_port }}
```

#### Linked Issues ####
N/A


dereknola (approved):

dereknola (changes requested):
You need to add the dependency for this collection to
https://github.com/k3s-io/k3s-ansible/blob/main/collections/requirements.yml

dereknola:
@sestegra you'll need to address the lint issues. Looking at CI, I'm not sure if the formatting of the requirements.yml is affecting the install of the utils collection during testing.

sestegra:
Workflows fixed now.

dereknola (approved):
The openrc error is unrelated to this PR.

Issue #551, opened 2026-08-04
Add variable to disable default firewall rules


mbathe19:
Hello,

when running `site.yml`, a task checks and applies the default UFW rules. While this is helpful for a basic setup, it can raise security concerns in more restrictive environments. For example, the etcd ports do not need to be accessible from everywhere; they should only be accessible from the master nodes.

To address this, I created my own UFW policies. However, when `site.yml` is executed again, these custom rules are overwritten by the default firewall rules.

I would suggest adding a variable to enable or disable the deployment of the default firewall rules.

Example:

```yaml
# k3s-ansible/roles/prereq/tasks/main.yml
- name: Allow UFW Exceptions
  when:
    - manage_firewall | default(true) | bool
    - ansible_facts.services['ufw'] is defined
    - ansible_facts.services['ufw'].state == 'running'
```

This could also be applied to the `firewalld` block.

Then the variable could be added to `inventory-sample.yml`:
```yaml
manage_firewall: false
```

I would be happy to open a PR for this if you are interested.

Regards,
Michel

dereknola:
PRs are appreciated.

PR #552, opened 2026-08-06
feat: add firewall variable

mbathe19:
#### Changes ####
- added `when` conditions to the firewall tasks.
- added the variable to `inventory-sample.yml`.

#### Linked Issues ####
- https://github.com/k3s-io/k3s-ansible/issues/551


roles/prereq/tasks/main.yml
```diff
@@ -45,6 +45,7 @@
 
 - name: Handle modern nftables/iptables-nft stack (Arch Linux 6.18+)
   when:
+    - manage_firewall | default(true) | bool
```
dereknola:
```suggestion
    - manage_firewall
```

Set manage_firewall in https://github.com/k3s-io/k3s-ansible/blob/main/roles/prereq/defaults/main.yml to default to `true`

roles/prereq/tasks/main.yml
```diff
@@ -72,6 +73,7 @@
 
     - name: Configure nftables include and K3s rules fragment
       when:
+        - manage_firewall | default(true) | bool
```
dereknola: Define default elsewhere

roles/prereq/tasks/main.yml
```diff
@@ -104,6 +106,7 @@
 
 - name: Allow UFW Exceptions
   when:
+    - manage_firewall | default(true) | bool
```
dereknola: Same comment as above

roles/prereq/tasks/main.yml
```diff
@@ -154,6 +157,7 @@
 
 - name: Allow Firewalld Exceptions
   when:
+    - manage_firewall | default(true) | bool
```
dereknola: Same comment as above

dereknola (changes requested):

dereknola:
@mbathe19 You'll need to address the lint issues.

PR #562, opened 2026-09-15
feat: roll the servers one at a time and wait for each to serve again

kristofer-atlas:
#### Changes ####

- The server and agent roles restart k3s on every run. On a cluster with three or more servers, a play that runs against the whole group restarts every etcd member at once and takes the quorum down with it. `k3s_server_serial` and `k3s_agent_serial` set the `serial` of the two plays so the role finishes on one node before the next starts. Both default to `100%`, the group in one batch, so a run that sets neither behaves as it does today. `serial` is a play keyword, templated before Ansible binds a host, so the value comes from `--extra-vars` or from the `vars` of the `import_playbook`.
- `k3s_wait_ready` holds a node until the API server reports it `Ready`. `k3s_server_wait_etcd_voters` holds a server until every server the run has started so far reports `EtcdIsVoter=True`, which `Ready` does not cover: a restarted member rejoins as a learner and counts towards no quorum until it catches up. Both default to `false`.
- The existing join barrier now counts the servers the run has started rather than the whole group, so a fresh HA install rolled one node at a time completes.

#### Linked Issues ####

- [k3s-io/k3s-ansible#485](https://github.com/k3s-io/k3s-ansible/issues/485)


dereknola:
How is this any better than just running the upgrade with `--forks=1`

kristofer-atlas:
`--forks=1` serialises task execution but not the rollout: every server still restarts within the same play run with no gate between them, so a slow restart on one node overlaps the next node's restart and etcd can lose quorum. The `serial: 1` play plus the Ready and etcd-voter waits make each node's return to service the precondition for touching the next one. This almost bit me recently.

Also, having to remember to run with a flag can be risky. Forgetting the flag could case whole cluster degradation. Codifying the rollout is a more robust approach.

PR #563, opened 2026-09-15
feat: add optional readiness gates to the server and agent roles

kristofer-atlas:
The roles restart k3s on every run and nothing holds a node until it serves again, so a run that rolls the servers one at a time starts the next restart while the previous member may still be coming back. On an embedded etcd cluster that is how a rolling run loses the quorum: a member that has just restarted rejoins as a learner and becomes a voter only once it has caught up, `Ready` does not report that, and a learner does not count towards the quorum the next restart needs. `k3s_wait_ready` holds each node at the end of its role until the API server reports it `Ready`, with the agent check delegated to the first server because an agent holds no kubeconfig, and `k3s_server_wait_etcd_voters` additionally holds each server until every control-plane node carries `EtcdIsVoter=True`. Both default to `false` and are skipped in check mode, so a run that sets neither behaves as it does today.


kristofer-atlas:
Superseded by https://github.com/k3s-io/k3s-ansible/pull/562, which carries the readiness gates together with the serial rollout they exist for.

